PR #13, opened 2017-05-04
Configurable example solution patern

### LINKED ISSUE #12: Allow tracks freedom to decide where example solutions should go
## Background

Among the [improvements](https://github.com/exercism/xruby/issues/485) the ruby track is making to its generator is moving all the pieces of a generator under the exercise's `.meta` directory. We'd like to move the example solution there, too (under `.meta/solutions`. Our track logic now supports having the example solution in that location, but we can't move ahead because configlet can't find the solutions 😭 

Rather than having all tracks adopt our (clearly optimal) new location for example solutions, it would be awesome if configlet allowed each track to choose an idiomatic location.

## Proposed Solution(s)

In the interests of flexibility, perhaps configlet could seperate out specifying what should be validated from doing the validation, so tracks have the option to override default validation as needed? 

Failing that, we could retain the `example.<ext>` naming scheme under `exercises/<slug>` but allow the file to be anywhere under `exercises/<slug>` and its subdirectories?

Insti:
Fixes: #12 

Previously the pattern used to check if an example solution was present for a file was hardcoded to `[Ee]xample`.

Since support was added for ignored `.meta` directory it is no longer necessary for a solution filename to contain `example` to avoid being served to the student.

This patch adds support for configuring the pattern used to detect the presence of an example solution.

It is set via the optional key `solution_pattern` in the track's `config.json` file. The value is a string which is used to create a regular expression.

If the key is not present the default value of `[Ee]xample` will be used.

Edit: fixed json key name

### DIFF
--- a/configlet/fixtures/track/config.json
+++ b/configlet/fixtures/track/config.json
@@ -3,6 +3,7 @@
   "language": "Gemstones",
   "repository": "https://github.com/exercism/xgemstones",
   "active": true,
+  "solution_pattern": "[Ee]xample|\\.meta/solutions/[^/]*\\.rb",
   "problems": [
     "amethyst",
     "beryl",
--- a/configlet/fixtures/track/config.json
+++ b/configlet/fixtures/track/config.json
@@ -8,7 +8,8 @@
     "beryl",
     "crystal",
     "melanite",
-    "sapphire"
+    "sapphire",
+    "rhodolite"
   ],
   "deprecated": [
     "opal",
--- a/configlet/config.go
+++ b/configlet/config.go
@@ -9,16 +9,17 @@ import (
 
 // Config is an Exercism track configuration.
 type Config struct {
-	path       string
-	Slug       string
-	Language   string
-	Active     bool
-	Repository string
-	Problems   []string
-	Exercises  []Exercise
-	Ignored    []string
-	Deprecated []string
-	Foregone   []string
+	path            string
+	Slug            string
+	Language        string
+	Active          bool
+	Repository      string
+	Problems        []string
+	Exercises       []Exercise
+	Ignored         []string
+	Deprecated      []string
+	Foregone        []string
+	SolutionPattern string `json:"solution_pattern"`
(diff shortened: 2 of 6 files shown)


kotp:
I believe this fixes#12 ...

kytrinyx:
Thanks @Insti, the approach you describe is much better.

I'll take a close look at this in the next couple of days.

configlet/fixtures/track/config.json
```diff
@@ -3,12 +3,14 @@
   "language": "Gemstones",
   "repository": "https://github.com/exercism/xgemstones",
   "active": true,
+  "solutionpattern": "[Ee]xample|\\.meta/solutions/[^/]*\\.rb",
```
kytrinyx: The key should be `solution_pattern` to match the style of the rest of the config (`test_pattern`, `ignore_pattern`).

kytrinyx (changes requested):
Overall this looks great. I just have the one request related to consistency in the config.json key name.

configlet/fixtures/track/config.json
Insti: Good point, I will change that when I get a chance in the next day or so. :+1:

Insti:
Done, do you want me to force push so the original commits use the underscore version?

kytrinyx:
> do you want me to force push so the original commits use the underscore version?

Yes, please!

Insti:
Done, should be ready to merge now.

kytrinyx (approved):

kytrinyx:
I've published a new release. The travis builds should pick it up automatically.

Insti:
Awesome. Thanks.

Insti:
Awesome. Thanks.

Insti:
Documented here: https://github.com/exercism/x-common/pull/778

Insti:
Confirmed working with travis via the xruby track.

kytrinyx:
:tada:

PR #19, opened 2017-06-11
Make the error messages more expressive

kytrinyx:
@stkent Would you take a look at the new error messages? Examples can be found in the evaluate_test.go file in this PR.

### DIFF
--- a/evaluate.go
+++ b/evaluate.go
@@ -27,23 +27,23 @@ func Evaluate(path string) bool {
 	configErrors := []ConfigError{
 		{
 			check: track.MissingProblems,
-			msg:   "-> No directory found for %v.\n",
+			msg:   "-> An exercise with slug '%v' is referenced in config.json, but no implementation was found.\n",
 		},
 		{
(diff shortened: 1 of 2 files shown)


stkent:
@kytrinyx brilliant! I'll update #20 in a sec to mirror this error language more closely.

evaluate.go
```diff
@@ -27,23 +27,23 @@ func Evaluate(path string) bool {
 	configErrors := []ConfigError{
 		{
 			check: track.MissingProblems,
-			msg:   "-> No directory found for %v.\n",
+			msg:   "-> An exercise named '%v' is referenced in config.json, but no implementation was found.\n",
```
stkent: Maybe `An exercise named` -> `An exercise with slug`?

stkent: Or just `'%v' is referenced in config.json`...?

kytrinyx: Oh, good point! An exercise with slug is much better. Let me change that.

kytrinyx: Fixed!

PR #20, opened 2017-06-11
Update configlet check descriptions

stkent:


README.md
```diff
@@ -7,12 +7,13 @@ certain assumptions. Configlet makes it simple to verify up-front
 that the changes to existing problems or the addition of new problems
 will play nicely with the API.
 
-There are three common problems that occur:
+Configlet checks for the following configuration issues:
```
stkent: Reworded to avoid using 'problems' in a second context.

stkent: Removed 'three' since there were 4 variants listed!

README.md
```diff
@@ -7,12 +7,13 @@ certain assumptions. Configlet makes it simple to verify up-front
...
 
-1. The `config.json` [(documented here)](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#track-configuration-file) might be invalid.
-1. Problems might be missing a reference solution.
-1. Problems might be implemented (test suite + reference solution), but not configured.
-1. Slugs in the configuration might not have a corresponding problem.
+1. `config.json` contents that are invalid according to [the specification](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#track-configuration-file).
+1. Inconsistencies between the list of declared problems and implementations of those problems:
+    * Problems that are declared in the configuration file but missing all implementation files.
+    * Problems that are declared in the configuration file but missing a reference solution.
+    * Problems that are fully implemented (test suite + reference solution are both present) but not declared in the configuration file.
```
stkent: Made language around the inconsistency cases more, ahem, consistent!

stkent:
Updated in light of #19 and https://github.com/exercism/docs/blob/master/glossary.md.

@kytrinyx is https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#track-configuration-file still the right target for this README? It refers to the `problems` key, ignored files, etc. Not sure if that file needs an update or if there's a newer location we should use instead in the docs repo?

kytrinyx:
Ah! No, let's point to the glossary rather than the x-common contributing file. Great find.

kytrinyx:
Wait, I see what you mean.

How about pointing to https://github.com/exercism/docs/issues/12 for now? I'll add a note to that one to update this README once that documentation exists.

kytrinyx:
This looks great, by the way. I think we can merge it as is, since we now have a note to update this when the documentation is updated.

stkent:
Perfect!

PR #24, opened 2017-07-09
Add generate command for exercise readmes

kytrinyx:
This builds on top of https://github.com/exercism/configlet/pull/23 and must not be merged until that has landed.

This assumes that every exercise in every track has a file `.meta/readme.go.tmpl`, which uses Go's text/template package.

It uses Go's text/template package. I've documented what I have so far [in the docs repository](https://github.com/exercism/docs/blob/master/maintaining-a-track/regenerating-exercise-readmes.md).

The TL;DR for why I think Go's text/template is a reasonable choice for us is:

1. Our shared tooling is already in Go (because it's easy to distribute / the binary has no dependencies)
1. It's fast.
1. We don't need anything complicated—just something that can interpolate values either conditionally or unconditionally. While `text/template` is a full-fledged templating language (and—as I've discovered—some people dislike it with a passion), we don't need any of the complex things.
1. Lastly, I'm mostly the one maintaining the tooling, so I reached for something that I am familiar with and comfortable with.

The following command will generate all the exercise READMEs for a track:

```
configlet generate path/to/track
```

If you only want to generate the readme for a single exercise, pass the flag `-only` with the slug.

```
configlet generate path/to/track --only bob
```

Given the following template, the README generated will be identical to the one that Trackler generates:

```
# {{ .Spec.Name }}

{{ .Spec.Description -}}
{{- with .Hints }}
{{ . }}
{{ end }}
{{- with .TrackInsert }}
{{ . }}
{{ end }}
{{- with .Spec.Credits -}}
## Source

{{ . }}
{{ end }}
## Submitting Incomplete Solutions
It's possible to submit an incomplete solution so you can see how others have completed the exercise.
```

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -1,0 +1,66 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/exercism/configlet/track"
+	"github.com/spf13/cobra"
+)
+
+var (
+	genSlug string
+)
+
+// generateCmd represents the generate command
+var generateCmd = &cobra.Command{
+	Use:   "generate",
+	Short: "Generate exercise READMEs for an Exercism language track",
+	Long: `Generate READMEs for Exercism exercises based on contents of
+	a number of different files.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			fmt.Fprintf(os.Stderr, "USAGE: configlet generate path/to/track")
+			os.Exit(1)
+		}
+
+		path, err := filepath.Abs(filepath.FromSlash(args[0]))
+		if err != nil {
+			fmt.Fprintf(os.Stderr, err.Error())
+			os.Exit(1)
+		}
+		root := filepath.Dir(path)
+		trackID := filepath.Base(path)
+
+		var exercises []track.Exercise
+
+		if genSlug != "" {
+			exercises = append(exercises, track.Exercise{Slug: genSlug})
+		} else {
+			track, err := track.New(path)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, err.Error())
+				os.Exit(1)
+			}
+			exercises = track.Exercises
+		}
+
+		errs := []error{}
+		for _, exercise := range exercises {
+			readme, err := track.NewExerciseReadme(root, trackID, exercise.Slug)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			if err := readme.Write(); err != nil {
+				errs = append(errs, err)
+			}
+		}
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(generateCmd)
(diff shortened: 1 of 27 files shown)


petertseng:
Hi.

Given that it will likely be the case that many exercises will have the same template, then might I suggest that support for a default template at a well-known location be provided? I'll just spitball with something like, say, `docs/default_exercise_template.go.tmpl`. My assumption is that for each track, many exercises will continue to have the same template. In only a few exercises per track will a different per-exercise template be needed.

So a default template will

1. reduce duplication
2. make the PRs easier to review because there will be ~half as many files added
3. make track-wide README updates in the future easier to enact, since only one file will need to be changed instead of a number scaling with the number of exercises

kytrinyx:
I like that suggestion a lot.

> My assumption is that for each track, many exercises will continue to have the same template.

A very sound assumption.

I don't think that the `docs` is the right place for it, necessarily, though on second thought I'm not sure why not. It feels kind of like a different thing.

kytrinyx:
I wonder if we should have a directory at the root of the repository for configuration things.

It would potentially contain both the `EXERCISE_README_INSERT.md` and the `readme.go.tmpl`.

At first I thought that if we have a default template we could get rid of the readme insert, but we would still need it for any exercises that override the default.

I originally also thought that we'd get rid of the `HINTS.md` file and just inline it into the template. But if we're using a default template, we'll still need HINTS.md. We should still move HINTS into the `.meta` directory, though.

petertseng:
> I wonder if we should have a directory at the root of the repository for configuration things.

Ah nice. That would restore the property "every file F.ext in docs corresponds to a file that appears in exercism.io/languages/$LANG/$F".

> At first I thought that if we have a default template we could get rid of the readme insert, but we would still need it for any exercises that override the default.

> I originally also thought that we'd get rid of the HINTS.md file and just inline it into the template. But if we're using a default template, we'll still need HINTS.md.

Ah, right. I guess keeping those files is the right balance, then. On the one hand there's a desire to have fewer files, but on the other hand given that there will presumably be exercises with non-default templates they will want to take advantage of the readme insert (deduplication).

> We should still move HINTS into the .meta directory, though.

It would be possible to take the opportunity to rename the file if desired. I have no opinion on the desirability.

kytrinyx:
> That would restore the property "every file F.ext in docs corresponds to a file that appears in exercism.io/languages/$LANG/$F".

Yeah. That's what I was thinking. It's been bugging me a bit ever since we made it that way. I just couldn't think of anything better at the time. *sigh* Sometimes I am a bit slow.

> It would be possible to take the opportunity to rename the [HINTS.md] file if desired.

Yeah, I don't have any preferences.

OK, given all of this, I'm going to open an issue in the discussions repository so we can get feedback on all of this.

Thanks for helping think it through.

kytrinyx:
I opened the issue here: https://github.com/exercism/discussions/issues/165

PR #25, opened 2017-07-14
minor update readme.md

rpottsoh:
Point link to problem-specifications instead of x-common

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ will play nicely with the API.
 
 Configlet checks for the following configuration issues:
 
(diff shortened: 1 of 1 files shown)


rpottsoh:
@kytrinyx should I also change the last line to read:
```bash
$ configlet lint .
```
instead of:
```bash
$ configlet .
```

kytrinyx:
Oh, that's a great point. Yes, please do!

kytrinyx:
Thank you @rpottsoh 🌷

Issue #8, opened 2016-11-07
Verify the presence of a test file


kytrinyx:
@behrtam commented in https://github.com/exercism/xpython/pull/392:

> ./bin/configlet only checks that for every exercises in the config.json there is also an example implementation and vice versa. It doesn't check if there is also a test file for every exercise.

We should actually read the `test_pattern` from the `config.json` file (defaulting to `/test/i` if `test_pattern` isn't defined) and verify this for each exercise.

nywilken:
@kytrinyx I would like to help out with this issue, if it hasn't already been taken. 

I'm working through the code changes and reading the [doc](https://github.com/exercism/docs/blob/9df310839d0d7413aec5915d13758c15c6a3388d/language-tracks/configuration/README.md) to get an understanding of things. Please let me know if there are any other threads I should read through. Thanks.

kytrinyx:
@nywilken that would be great. I don't _think_ there's any more context you need than what is here, but if you get stuck or confused don't hesitate to ask.

nywilken:
@kytrinyx sounds good, thanks.

kytrinyx:
@nywilken I've sent you an invite to the Exercism organization. If/when you accept it, drop a line here so I can assign this issue to you.

nywilken:
@kytrinyx invite accepted, thanks.

kytrinyx:
Welcome ✨ 💛 🌻 💛 ✨

nywilken:
@kytrinyx thanks for the encouraging welcome. I submitted PR #32, as a first pass, to address this issue. I left a few assumptions and remaining tasks in the comments of the PR. Looking forward to the review.

kytrinyx:
Awesome, thanks so much for that!

nywilken:
Closed by #32

rpottsoh:
Looks like this should have closed when #32 was merged.

PR #32, opened 2017-07-25
configlet: Verify presence of test file

nywilken:
Address Configlet #8, by providing a mechanism for reading test_pattern from the config.json file (defaulting to /test/i if test_pattern isn't defined) and verifying that each exercise has an accompanying test file that matches the provided test_pattern.

Assumptions:
- Changing the signature for excercise.New is not an option; nor is it necessary. 
- Track initialization does not have a hard dependency on the existence of a test file. In other words, it is safe to ignore any errors returned by excercise. LoadTestSuitePath in track.New(). 

Remaining Tasks:
- [x] Add test cases to exercise_test
- [x] Test lint on various tracks (Go, Rust, Erlang, Python, JavaScript)
- [x] Squash commits
- [x] Update config.json [configuration documentation](https://github.com/exercism/docs/tree/master/language-tracks/configuration) with link to supported regex [syntax](https://golang.org/pkg/regexp/syntax/) (See https://github.com/exercism/docs/pull/78)

Superscedes https://github.com/nywilken/exer-configlet/pull/1

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -56,6 +56,10 @@ func lintTrack(path string) bool {
 			msg:   "-> The implementation for '%v' is missing an example solution.\n",
 		},
 		{
+			check: missingTestSuite,
+			msg:   "-> The implementation for '%v' is missing a test suite.\n",
+		},
+		{
 			check: foregoneViolations,
 			msg:   "-> An implementation for '%v' was found, but config.json specifies that it should be foregone (not implemented).\n",
 		},
@@ -139,6 +143,25 @@ func missingSolution(t track.Track) []string {
 
 	slugs := []string{}
 	for slug, ok := range solutions {
+		if !ok {
+			slugs = append(slugs, slug)
+		}
+	}
+	return slugs
+}
+
+func missingTestSuite(t track.Track) []string {
+	tests := map[string]bool{}
+	for _, exercise := range t.Exercises {
+		tests[exercise.Slug] = exercise.HasTestSuite()
+	}
+	// Don't complain about missing test suite in foregone exercises.
+	for _, slug := range t.Config.ForegoneSlugs {
+		tests[slug] = true
+	}
+
+	slugs := []string{}
+	for slug, ok := range tests {
 		if !ok {
 			slugs = append(slugs, slug)
 		}
--- a/track/config.go
+++ b/track/config.go
@@ -6,6 +6,12 @@ import (
 	"io/ioutil"
 )
 
+// PatternGroup holds matching patterns defined in an Exercism track configuration.
+type PatternGroup struct {
+	SolutionPattern string `json:"solution_pattern"`
+	TestPattern     string `json:"test_pattern"`
+}
+
 // Config is an Exercism track configuration.
 type Config struct {
 	Language        string
@@ -13,15 +19,19 @@ type Config struct {
 	Exercises       []ExerciseMetadata
 	DeprecatedSlugs []string `json:"deprecated"`
 	ForegoneSlugs   []string `json:"foregone"`
-	SolutionPattern string   `json:"solution_pattern"`
+	PatternGroup
 }
 
 // NewConfig loads a track configuration file.
-// The config has a default solution pattern if none is provided in the file.
-// The solution pattern is sued to determine if an exercise has a sample solution.
+// The config has a default solution and test pattern if not provided in the file.
+// The solution pattern is used to determine if an exercise has a sample solution.
+// The test pattern is used to determine if an exercise has a test suite.
 func NewConfig(path string) (Config, error) {
 	c := Config{
-		SolutionPattern: "[Ee]xample",
+		PatternGroup: PatternGroup{
+			SolutionPattern: "[Ee]xample",
+			TestPattern:     "(?i)test",
+		},
 	}
 
 	bytes, err := ioutil.ReadFile(path)
--- a/track/config_test.go
+++ b/track/config_test.go
@@ -25,3 +25,9 @@ func TestDefaultSolutionPattern(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, "[Ee]xample", c.SolutionPattern)
 }
+
+func TestDefaultTestPattern(t *testing.T) {
+	c, err := NewConfig("../fixtures/empty.json")
+	assert.NoError(t, err)
+	assert.Equal(t, "(?i)test", c.TestPattern)
+}
--- a/track/exercise.go
+++ b/track/exercise.go
@@ -10,14 +10,41 @@ import (
 
 // Exercise is an implementation of an Exercism exercise.
 type Exercise struct {
-	Slug         string
-	SolutionPath string
+	Slug          string
+	SolutionPath  string
+	TestSuitePath string
 }
 
 // NewExercise loads an exercise.
-func NewExercise(root string, rgx *regexp.Regexp) (Exercise, error) {
+func NewExercise(root string, pg PatternGroup) (Exercise, error) {
 	ex := Exercise{
 		Slug: filepath.Base(root),
+	}
+
+	err := setPath(root, pg.SolutionPattern, &ex.SolutionPath)
+	if err != nil {
+		return ex, err
+	}
+
+	err = setPath(root, pg.TestPattern, &ex.TestSuitePath)
+	if err != nil {
+		return ex, err
+	}
+
+	return ex, err
+}
+
+// setPath sets the value of field to the file path matched by pattern.
+// The resulting file path, if matched, will be relative to root.
+func setPath(root, pattern string, field *string) error {
+
+	if pattern == "" {
+		return nil
+	}
+
+	rgx, err := regexp.Compile(pattern)
+	if err != nil {
+		return err
 	}
 
 	walkFn := func(path string, info os.FileInfo, err error) error {
@@ -27,13 +54,17 @@ func NewExercise(root string, rgx *regexp.Regexp) (Exercise, error) {
 
 		if rgx.Match([]byte(path)) {
 			prefix := fmt.Sprintf("%s%s", root, string(filepath.Separator))
-			ex.SolutionPath = strings.Replace(path, prefix, "", 1)
+			*field = strings.Replace(path, prefix, "", 1)
 		}
 		return nil
 	}
 
-	err := filepath.Walk(root, walkFn)
-	return ex, err
+	return filepath.Walk(root, walkFn)
+}
+
+// HasTestSuite checks that an exercise has a test suite.
+func (ex Exercise) HasTestSuite() bool {
+	return ex.TestSuitePath != ""
 }
 
 // IsValid checks that an exercise has a sample solution.
--- a/track/exercise_test.go
+++ b/track/exercise_test.go
@@ -2,7 +2,6 @@ package track
 
 import (
 	"path/filepath"
-	"regexp"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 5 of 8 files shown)


cmd/lint.go
```diff
@@ -56,6 +56,10 @@ func lintTrack(path string) bool {
 			msg:   "-> The implementation for '%v' is missing an example solution.\n",
 		},
 		{
+			check: missingTestSuite,
+			msg:   "-> The implementation for '%v' is missing an accompanying test suite.\n",
```
kytrinyx: I would remove `accompanying`, since the test suite _is_ essentially the implementation.

track/config_test.go
```diff
@@ -25,3 +25,9 @@ func TestDefaultSolutionPattern(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, "[Ee]xample", c.SolutionPattern)
 }
+
+func TestDefaultTestPattern(t *testing.T) {
```
kytrinyx, line 29: I like this test, thank you. It's very easy to understand exactly what is going on.

track/exercise.go
```diff
@@ -20,20 +21,36 @@ func NewExercise(root string, rgx *regexp.Regexp) (Exercise, error) {
 		Slug: filepath.Base(root),
 	}
 
-	walkFn := func(path string, info os.FileInfo, err error) error {
+	err := filepath.Walk(root, fieldPath(root, rgx, &ex.SolutionPath))
+	return ex, err
+}
+
+// LoadTestSuitePath loads the test suite path for exercise using the provided file pattern.
+func (ex *Exercise) LoadTestSuitePath(root string, rgx *regexp.Regexp) error {
```
kytrinyx: Verifying the presence of the solution and the test suite are essentially the same thing twice, and I think it is important to find a way to treat them in the same way.

Should we pass the `config` value to the new exercise so that we have both of the regexes?

track/exercise_test.go
```diff
@@ -63,3 +63,43 @@ func TestExerciseSolutionPaths(t *testing.T) {
...
+			pattern:   "secret",
+			testsuite: "subdir/.secret-solution.ext",
+		},
+	}
+
+	for _, test := range tests {
+		path := filepath.FromSlash("../fixtures/fake-exercise")
+
+		rgx, err := regexp.Compile("[Ee]xample")
+		assert.NoError(t, err)
```
kytrinyx: Since this doesn't change from test to test, it can be done outside the loop.

track/exercise_test.go
```diff
@@ -63,3 +63,43 @@ func TestExerciseSolutionPaths(t *testing.T) {
...
+		},
+		{
+			// It finds files in a subdirectory.
+			pattern:   "secret",
+			testsuite: "subdir/.secret-solution.ext",
+		},
+	}
+
+	for _, test := range tests {
+		path := filepath.FromSlash("../fixtures/fake-exercise")
```
kytrinyx: This is the same each time, so it should probably live outside the loop.

kytrinyx (changes requested):
Overall this looks great. I have a few minor comments, and one bigger design question that I'd like to resolve before bringing this in.

track/config_test.go
nywilken: Awesome, glad to hear that.

track/exercise.go
nywilken: I agree 💯 on the presence of the solution and test suite being the same thing. I sort of made the assumption that changing the signature of the constructor would be an unwanted change seeing as it would be a breaking change. This probably would've been a good first question.

TBH, I think changing Exercise to support multiple regex patterns would clean and simplify things. I like the idea of passing config values; are you thinking the entire config value or just the patterns (e.g new type)? Seeing as not all Config data pertains  to an Exercise.  

I originally thought about passing in a map of patterns to the constructor which we could then compile and use to set the appropriate fields. But in thinking about it we might be better off creating a Patterns type. Thoughts?

nywilken:
@kytrinyx I updated the code base with your recommendations and left my thoughts on the larger design question. Thanks for the first review.

cmd/lint.go
```diff
@@ -151,17 +151,17 @@ func missingSolution(t track.Track) []string {
 }
 
 func missingTestSuite(t track.Track) []string {
-	testsuite := map[string]bool{}
+	testsuites := map[string]bool{}
```
kytrinyx: I missed this the first time around, but can we make this `testSuites` (mixedCase)?

track/exercise_test.go
```diff
@@ -63,3 +63,41 @@ func TestExerciseSolutionPaths(t *testing.T) {
...
+		},
+		{
+			// It finds files with an exact match.
+			pattern:   "[^_]+_test.ext$",
+			testsuite: "fake_test.ext",
+		},
+		{
+			// It finds files in a subdirectory.
+			pattern:   "secret",
+			testsuite: "subdir/.secret-solution.ext",
```
kytrinyx: I don't think we can serve hidden tests—that wouldn't make much sense.

track/exercise_test.go
```diff
@@ -63,3 +63,41 @@ func TestExerciseSolutionPaths(t *testing.T) {
 		assert.Equal(t, test.solution, ex.SolutionPath)
 	}
 }
+func TestExerciseLoadTestSuitePath(t *testing.T) {
+	tests := []struct {
+		pattern   string
+		testsuite string
```
kytrinyx: Here, too, I would prefer `testSuite` (or even just `path`—either would work).

track/exercise.go
kytrinyx: >  I sort of made the assumption that changing the signature of the constructor would be an unwanted change seeing as it would be a breaking change.

It's not a breaking change, since the only consumer of this code is the lint command, and that would get updated at the same time.

> are you thinking the entire config value or just the patterns

At first I was thinking a new type, but then I felt lazy and figured maybe we'd just pass the whole config... but you're right. Most of the config is _not_ relevant.

Let's do the clean thing and make a new type!

track/exercise_test.go
kytrinyx: Oh, rereading this I notice that the test isn't about the hidden thing. I think it's confusing because it's a hidden file and called `solution`. Let's add another fixture file that is named with `suite` or something (not test, but also not secret).

kytrinyx:
This is looking great. I'm really enjoying reviewing this—your code is really straight forward and easy to follow.

nywilken:
I can say the same - the code base is well thought-out and nicely documented 👍 
I'll work on updating the sections you called out; and will definitely improve the tests so that they make sense.

Regarding the new type any thoughts on where it should live? Outside of Config (e.g config_patterns.go)? Or within config.go? I'm was leaning towards the latter seeing as the type would never, outside of testing, be used on its own.  But I can see some advantages to having it within its own file. 

Thanks again for the help and guidance.

kytrinyx:
> Regarding the new type any thoughts on where it should live?

I think it makes sense to keep it in `config.go`. If we ever need it outside of the context of the config, we can move it later.

nywilken:
@kytrinyx I pushed out the latest refactor which includes a new FilePatterns type for your review. LMK your thoughts. 

If it's good to go I would like to squash the commits be for it gets merged. In the meantime, I'll will work on opening a PR for the docs. Thanks!

track/config.go
```diff
@@ -9,22 +9,32 @@ import (
 
 var errInvalidConfig = errors.New("invalid config file - try jsonlint.com")
 
+// FilePatterns is a collection of file matching patterns defined in an Exercism track configuration.
+type FilePatterns struct {
+	SolutionPattern string `json:"solution_pattern"`
+	TestPattern     string `json:"test_pattern"`
+}
```
kytrinyx: Overall I like the `FilePatterns` type. I think that helps clean the code up and keep things symmetrical. Embedding it in the Config works really well, too. Very nice.

I'm thinking about names. `FilePatterns.SolutionPattern` has a bit of an echo with the patterns/patterns. The variable name elsewhere is `patterns`, which makes it even more obvious.

A second thing about names is that `Patterns` makes me think that it's something that I could iterate over, like a map or a slice.

I don't want to throw a `Config` on there (`FilePatternConfig`), because we're sticking it _in_ the config, which would give us another echo.

I can't find a way around the plural, so I think we need to live with that one.

My suggestion for the echo is to take off `Pattern` from the individual fields: `patterns.Solution` and `patterns.Test`.

Thoughts?

As you can tell, I spend far too much agonizing over names in my life :-)

track/exercise_test.go
```diff
@@ -11,55 +10,101 @@ import (
...
+		patterns FilePatterns
 		solution string
-		pattern  string
 	}{
 		{
 			// It finds files in the root of the exercise directory.
-			pattern:  "[Ee]xample",
+			patterns: FilePatterns{
+				SolutionPattern: "[Ee]xample",
+				TestPattern:     "(?i)test",
```
kytrinyx: Does this blow up if you leave the TestPattern empty?

My thinking is this: anything that is laid out explicitly in test setup feels like it should be germane to the thing under test. TestPattern is incidental to testing the solution pattern, so this data can seem a bit distracting here.

My suggestion would be to either leave it blank (if the code doesn't blow up), or (if it does blow up), to explicitly set it once inside the loop rather than as part of the test case definition.

Does that make sense? if not let me know and I'll try to articulate this more clearly.

track/exercise_test.go
```diff
@@ -11,55 +10,101 @@ import (
...
+}
+func TestExerciseTestSuitePaths(t *testing.T) {
+	tests := []struct {
+		patterns FilePatterns
+		path     string
+	}{
+		{
+			// It finds files in the root of the exercise directory.
+			patterns: FilePatterns{
+				SolutionPattern: "[Ee]xample",
```
kytrinyx: Ditto here. If we can make it so that it doesn't look like the `SolutionPattern` is important to this test, that would make it easier for people to read/understand later.

kytrinyx:
Sweet. I think this is almost there. A couple of small stylistic/philosophical comments, and then we should be ready to squash and merge.

track/config.go
nywilken: I'm right there with you. I spent some time thinking about the name, myself, as I wasn't reallying feeling the whole FilePatterns name.

So I'm happy you called that out; and I totally see the point on how it reads and how the 's' can imply an iterateble type. 

After a little more thinking, I came up with a new name: `config.PaterrnGroup`

Which I think reads much better as `config.PatternGroup.SolutionPattern`

What do you think? 🤞

>My suggestion for the echo is to take off Pattern from the individual fields: patterns.Solution and patterns.Test.

For the record, I am really trying to keep SolutionPattern and TestPattern as the field names because with the auto promotion `config.SolutionPattern` would be clearer than `config.Solution`

track/exercise_test.go
nywilken: Makes complete sense. Really nice way of looking at things. Thanks for that lesson.

>Does this blow up if you leave the TestPattern empty?

It doesn't blow up but it does introduce an expected side effect where it matches any file when given an empty string. I will add some logic to address this and will add a test for an empty `PatternGroup`

track/config.go
kytrinyx: Ahhh, yes the auto-promotion is much cleaner in this way. Such a great point.

I think `PatternGroup` works. It still has a little bit of echo, but I think the trade-offs are worth it.

track/exercise_test.go
kytrinyx: I don't think we need a test for the empty pattern group, though, since when we load up a new config we explicitly set defaults. Since we always put the default in "real life" (so to speak), I think that it's fine to simply leave it off in this test, since the test is not using it at all.

nywilken:
Updated and squashed!

Let me know if I left any comments on concerns unaddressed.

track/config.go
```diff
@@ -9,22 +9,33 @@ import (
 
 var errInvalidConfig = errors.New("invalid config file - try jsonlint.com")
 
+// PatternGroup holds matching patterns defined in an Exercism track configuration.
+// See https://github.com/exercism/docs/tree/master/language-tracks/configuration for more information on matching patterns.
```
kytrinyx: Is this going to have more information? I'd rather not point to it (at least not yet), since I'm not sure that it will actually help.

kytrinyx:
This looks great. I have one final question about the documentation link, and this also has a conflict against the current master.

nywilken:
Thanks for looking over the final changes. I resolved the conflict on master so we should be good to go. 

You are right the link doesn't really provide more information - link removed.

kytrinyx (approved):

kytrinyx:
Lovely. Let's 🚢 this!

Issue #18, opened 2017-06-10
Investigate: what would it look like to lint the JSON format of config.json?


kytrinyx:
We want to normalize the JSON so that we can script changes to it without adding a lot of whitespace noise.

robphoenix:
So far I've not found a JSON linter written in Go. I did wonder if we could just re-marshall the verified Config using `MarshallIndent`, see https://github.com/robphoenix/configlet/commit/de28dc124b10a462db3013799bf6d93b7505af1f, and then re-write the output back to file, but of course it wouldn't be included in the commit, would it? And it wouldn't output any response indicating whitespace inconsistencies travis could fail on.

Will this require writing a JSON linter in Go, or will it just need to check that each line is appropriately indented? How full featured does it need to be? How much is taken care of by unmarshalling the file into the appropriate struct?

kytrinyx:
We could add a separate command to `configlet` that you can run locally before committing (kind of like `gofmt` but for `config.json`)

robphoenix:
That would work. Would it be run by maintainers every now and then, or suggested to contributors in track README's perhaps? I'll send a PR later this week, or would it better to wait for #23 (which I've been meaning to properly read through, having been wrestling with Cobra myself recently)

kytrinyx:
I think that if we use it, we should suggest that it be used by both contributors and maintainers whenever the config.json gets changed.

We should totally wait for #23 to land. I'm waiting on one last PR to be merged—all the 60+ other PRs that this one is relying on have been merged. I'll ping them (again).

robphoenix:
Grand, I'll familiarize myself with #23 and work on a PR to implement what I've outlined above.

kytrinyx:
Alrighty, #23 has landed!

You should be able to generate a new command with `cobra`, e.g. `cobra add fmt`, which will add a new file `cmd/fmt.go`, which in turn creates a new command `configlet fmt`. (Delete all the extraneous comments, license annotations, etc, they're just a distraction).

kytrinyx:
Oh, also: you can load up a config file with `track.NewConfig(pathToTrack)`. Check out https://github.com/exercism/configlet/blob/master/track/config_test.go

robphoenix:
Cool, thanks @kytrinyx I can't get to this until Tuesday at the earliest, but I'll aim to get something here for review next week.

kytrinyx:
@robphoenix that sounds great, thank you!

PR #34, opened 2017-07-27
Add fmt command

robphoenix:
This is for initial review, in response to #18


cmd/fmt.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func format(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
+		fmt.Fprintln(os.Stderr, "USAGE: configlet fmt path/to/track")
+		os.Exit(1)
+	}
+
+	path := args[0]
+	configFile := fmt.Sprintf("%s/config.json", path)
```
nywilken: When it comes to building file paths the code should be separator agnostic.The filepath pkg is perfect for building paths with the  correct separator on *nix and Windows.

cmd/fmt.go
```diff
@@ -0,0 +1,65 @@
...
+
+	err = ioutil.WriteFile(configFile, formatted, 0644)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+	fmt.Printf("formatted: %s\n", configFile)
+}
+
+func formatConfig(f string) ([]byte, error) {
```
nywilken: This looks like a good candidate to be a method of Config seeing as it is specific to the config type. It would also keep the json marshaling in one place.

nywilken:
@robphoenix the fmt command is off to a good start. Thanks!

I took a quick look at the initial submit and left a couple of comments that would be helpful to this PR.

cmd/fmt.go
```diff
@@ -0,0 +1,65 @@
...
+Call fmt command with path to track:
+
+  configlet fmt path/to/track
+`,
+	Run: format,
+}
+
+func format(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
+		fmt.Fprintln(os.Stderr, "USAGE: configlet fmt path/to/track")
```
kytrinyx: How would you feel about trying out the usage function? Not sure if it would help in any way.
https://github.com/spf13/cobra/blob/f20b4e9c32bb3e9d44773ca208db814f24dcd21b/command.go#L189

cmd/fmt.go
kytrinyx: We're using `filepath.Join` throughout the codebase for these.

cmd/fmt.go
```diff
@@ -0,0 +1,65 @@
...
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+
+	err = ioutil.WriteFile(configFile, formatted, 0644)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+	fmt.Printf("formatted: %s\n", configFile)
```
kytrinyx: Is there a diff format we could use, so that it only outputs the things that changed?

I think that the regular `gofmt` command has a diff flag, so there might be code in the standard library that does diffing nicely.

cmd/fmt_example_test.go
```diff
@@ -0,0 +1,61 @@
+package cmd
+
+import "fmt"
+
+func ExampleFormat() {
+	cfg, err := formatConfig("../fixtures/numbers/config.json")
```
kytrinyx: I'd use the `filepath.FromSlash()` here.

cmd/fmt_example_test.go
```diff
@@ -0,0 +1,61 @@
...
+	//     }
+	//   ],
+	//   "deprecated": [
+	//     "three"
+	//   ]
+	// }
+	// <nil>
+}
+
+func ExampleBrokenFormat() {
```
kytrinyx: Nice, thanks for adding a test for the broken one, too.

kytrinyx (changes requested):
This looks great. I have a couple of tiny nitpicks (and generally agree with the suggestion to move the writing logic onto the Config type).

cmd/fmt.go
robphoenix: I've added in the `SetUsageFunc`, though it currently doesn't add much currently. Looking at how other projects have used it, I think it will prove more useful when used with a template across all the commands.

cmd/fmt.go
robphoenix: So, `gofmt` calls out to the unix `diff` tool: https://github.com/golang/go/blob/master/src/cmd/gofmt/gofmt.go#L244

While Linux, and I assume macOS, will have `diff` pre-installed, Windows users won't, which isn't great, https://github.com/sqs/goreturns/issues/28#issuecomment-254603819
Windows users can install the diff tool, or there is an equivalent tool, called [FC (file compare)](https://www.howtogeek.com/206123/how-to-use-fc-file-compare-from-the-windows-command-prompt/) and so it could be possible to check the host OS and call out to each tool depending on what the OS is.
Alternatively I found a couple of diff libraries. [go-diff](https://github.com/sergi/go-diff) & [go-difflib](https://github.com/pmezard/go-difflib). `go-difflib` I've used here, even though it's not currently maintained, while `go-diff` is. The output from `go-diff` wasn't great, it was either a malformed output similar to `diff`, or a coloured string, highlighting changes, though the whole file would be outputted even if there are no changes, which is not particularly helpful, whereas the output fro `go-difflib` appears much like the output from `diff`.
I'm going to have a look into using `diff`/`FC`, but wanted to put this out there to get your thoughts on the options availble.

cmd/fmt.go
robphoenix: That makes a lot of sense, thanks @nywilken

cmd/fmt.go
kytrinyx: >  Windows users won't, which isn't great

Yeah, I feel like anything we do that would make Windows users feel like second-class citizens, or that we haven't considered their needs would be not great.

I am all for going with with the unmaintained library if it does what we need. If you have the cycles to look into `diff/FC` that would be great, but I don't want to put an extra burden on you!

robphoenix:
I'm afraid I won't get to the failing test until tomorrow.

kytrinyx:
It's all good!

I can't make a new release until a couple of tracks update their configs, so you're not holding anything up.

kytrinyx:
(and even then... we could always make another release—these ones are easy and quick)

cmd/fmt_example_test.go
robphoenix: Now that the `formatConfig` func has been moved to the `Write` method the broken config is only handled by the `NewConfig` func which was previously wrapped by `formatConfig` so as far as I can see this test is pretty much redundant. I'm not sure if this is an indication that there's something wrong with the `Write` method?

kytrinyx: No, I think the code looks fine, and agree that this test is redundant. I'm sure a year from now I'll think of something I'd want to move around... I'm a bit slow that way :)

track/config.go
```diff
@@ -14,12 +14,16 @@ type PatternGroup struct {
...
-	Active          bool
-	Exercises       []ExerciseMetadata
-	DeprecatedSlugs []string `json:"deprecated"`
-	ForegoneSlugs   []string `json:"foregone"`
+	Slug            string `json:"slug"`
+	Language        string `json:"language"`
+	Repository      string `json:"repository"`
+	Active          bool   `json:"active"`
+	IgnorePattern   string `json:"ignore_pattern"`
+	SolutionPattern string `json:"solution_pattern,omitempty"`
```
nywilken: This field has been moved to the PatternGroup type. Please see [line #11](https://github.com/exercism/configlet/pull/34/files?diff=unified#diff-558b703b03397a6515a9f2eff2efcc12R11)

The tag for TestPattern should also get changed to `json:test_pattern,omitempty`

track/config.go
```diff
@@ -14,12 +14,16 @@ type PatternGroup struct {
...
-	Language        string
-	Active          bool
-	Exercises       []ExerciseMetadata
-	DeprecatedSlugs []string `json:"deprecated"`
-	ForegoneSlugs   []string `json:"foregone"`
+	Slug            string `json:"slug"`
+	Language        string `json:"language"`
+	Repository      string `json:"repository"`
+	Active          bool   `json:"active"`
+	IgnorePattern   string `json:"ignore_pattern"`
```
nywilken: This field should probably be moved to the PatternGroup type as well. @kytrinyx can I get your thoughts on this, please. IgnorePattern is not really used by the Exercise type but it is a pattern string.

cmd/fmt.go
```diff
@@ -0,0 +1,81 @@
...
+func format(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
+		usageFunc(cmd)
+		return
+	}
+
+	path := args[0]
+	configFile := filepath.Join(path, "config.json")
+
+	original, err := ioutil.ReadFile(configFile)
```
nywilken: Wondering if we could get rid of this and just use track.NewConfig like you do on line 41. Then implement the String method for Config to return a non formatted json string. 

```
func (c Config) String() string {
  o, _ := json.Marshal(c)
  return string(o)
}
```

Which could then be used for the diffing on line 53

```
original := fmt.Sprint(cfg)
delta, err := diff([]byte(original), formatted)
```

Thoughts?

cmd/fmt_example_test.go
robphoenix: Ha, yeah, a little distance and a little time can provide a whole new perspective.

cmd/fmt.go
```diff
@@ -0,0 +1,81 @@
...
+		return
+	}
+	if delta == "" {
+		fmt.Println("No differences encountered.")
+		return
+	}
+	fmt.Printf("Changes made:\n\n%s", delta)
+}
+
+func diff(a, b []byte) (string, error) {
```
nywilken: If my suggestion about implementing the String method for Config sounds good to you we could change the signature of `diff` to take in a string for a and b. Which would mean that the Write method would have to return a string. Thoughts?

nywilken (changes requested):
@robphoenix this is looking really good. My changes to master for verifying a valid test file introduced a new type called `PatternGroup`, which now holds the fields SolutionPattern and TestPattern.

The latest commit is overriding SolutionPattern which is probably why things are failing. Please check out my comments on the requested changes. 

Also I think we might be able to cut down on the two file reads in the fmt command by implementing the Stringer interface for Config. I left some comments for you on that as well.

cmd/fmt.go
robphoenix: I'm going to stick with the unmaintained library for now, unless I find a better one. The problem with using `FC` would be that it stands for a completely different tool in Powershell, so `FC` is fine for `cmd.exe` but for people like me who use Powershell as their Windows terminal it won't work and then leads down a path of trying to cover every edge case.

robphoenix:
For reference this is the output when running `configlet fmt` on the Go track.

```
gopath/src/github.com/exercism/configlet [fmt ≡ +1 ~2 -0] 
> go run main.go fmt ~/Dropbox/code/gopath/src/github.com/robphoenix/exercism-go
Changes made:

@@ -1,0 +2 @@
+  "slug": "",
@@ -2,0 +4 @@
+  "repository": "",
@@ -4,0 +7,2 @@
+  "solution_pattern": "[Ee]xample",
+  "test_pattern": "(?i)test",
@@ -477,2 +481,2 @@
-          "Maps",
-          "Transforming"
+        "Maps",
+        "Transforming"
@@ -594,3 +598 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -604,3 +606 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -614,3 +614 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -636,3 +634 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -646,3 +642 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -656,3 +650 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -666,3 +658 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -676,3 +666 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -686,3 +674 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -696,3 +682 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -706,3 +690 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -716,3 +698 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -726,3 +706 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -736,3 +714 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -746,3 +722 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -756,3 +730 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -766,3 +738 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -776,3 +746 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -786,3 +754 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -796,3 +762 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -806,3 +770 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -816,3 +778 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -826,3 +786 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -836,3 +794 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -846,3 +802 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -856,3 +810 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -866,3 +818 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -876,3 +826 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -886,3 +834 @@
-      "topics": [
-
-      ]
+      "topics": []
@@ -930,0 +877,4 @@
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
@@ -935,0 +886,4 @@
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
@@ -940,0 +895,4 @@
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
@@ -945,0 +904,4 @@
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
@@ -950,0 +913,4 @@
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
@@ -955 +920,0 @@
-
```

robphoenix:
I notice that in the Go `config.json` the `slug` and `repository` fields are missing and so when formatted they are added with empty strings as values. Is this going to override any default behaviour?
If we add `omitempty` to the JSON tags then they won't be added to the `config.json` file. I'm nor sure what would be preferable.

cmd/fmt.go
robphoenix: Yeah, I think that makes a lot of sense. I didn't like the `ioutil.ReadFile`, feels really clunky.

cmd/fmt.go
robphoenix: It has to be cast to a string eventually, sounds good.

cmd/fmt.go
robphoenix: Actually, I've just remembered why I read the file rather than marshal the config. The diff needs to be between the original file and the newly formatted file. The output from marshalling the config is not necessarily going to be the same as the original file. For example if some crazy maintainers liked to have their `config.json` minified then the output of a string method that marshals the unmarshalled JSON is going to very different to the content of the minified `config.json`.

Maybe we could store the original file contents in the Config struct during `NewConfig`? And call them in the `String` method?

```
func NewConfig(path string) (Config, error) {
    content, _ := ioutil.ReadFile(path)
	c := Config{
		PatternGroup: PatternGroup{
			SolutionPattern: "[Ee]xample",
			TestPattern:     "(?i)test",
            fileContents: content,
		},
	}
```

```
func (c Config) String() string {
  return string(c.fileContents)
}
```

robphoenix:
@nywilken Thanks!

> My changes to master for verifying a valid test file introduced a new type called PatternGroup, which now holds the fields SolutionPattern and TestPattern.

Yeah, this stumped for a bit but I nailed it down eventually, gave me some debugging practice! :smile: 

> Also I think we might be able to cut down on the two file reads in the fmt command by implementing the Stringer interface for Config. I left some comments for you on that as well.

I've responded to your comments. I'm not sure this would output an accurate diff, as it wouldn't be using the content from the original file, rather the marshalled `Config` struct created from the file.

cmd/fmt.go
nywilken: Yeah, that makes sense. In fact even if the file is properly formatted, the fact that we call .Marshal in String() and .MarshalIndent in Write() there will always be a diff, I think. So it looks like we have to store the original contents to diff on formatting. 

> Maybe we could store the original contents  in the Config strict.

The original contents is sort of ephemeral so wouldn't want to add to Config. It would also introduce a case where we would have to check if c.fileContents is empty, for Config objects created as literals,  and return a Marshalled string of Config.

robphoenix: > The original contents is sort of ephemeral so wouldn't want to add to Config.

Makes sense.

I'll leave this as-is for now then.

robphoenix:
At some point this week I'll add to this so it formats the maintainers file also.

kytrinyx:
> I notice that in the Go config.json the slug and repository fields are missing and so when formatted they are added with empty strings as values. Is this going to override any default behaviour?

We're removing the slug and repository fields across all of Exercism. The slug is never used, and the repository can be inferred from the track ID.

> At some point this week I'll add to this so it formats the maintainers file also.

That is great!

cmd/fmt.go
nywilken: So I thought a little more about this issue and started to wonder if maybe we're trying to solve the wrong problem. More specifically, if instead of deserializing the JSON into a Config object then calling Write to format and save, we dropped the Config object entirely and just use the Encoder and Decoder types. 

Something like so:

Use a json decoder to decode the file into a map.
Create a new file using an encoder with the correct formatting.
Use the diff library to diff the decoded file against the newly encoded file (a little bit of work involved here). 
Then overwrite the original file with the new encoded file. 

Essentially making the fmt command a raw JSON formatter for any json file. 

Here's a quick example: https://play.golang.org/p/w8aSEZuyIH

Of course this would only work if we are okay with not actually loading the json file into a Config type, the encoder/decoders outputs are consistent and if the type map[string]interface{} can actually handle any json file structure. 

Thoughts?

Sorry for turning this PR into such a long commentary, this is just an interesting problem

kytrinyx: > Essentially making the fmt command a raw JSON formatter for any json file.

I like that idea—then the same encoder/decoder would work for the maintainers file, too.

> Of course this would only work if we are okay with not actually loading the json file into a Config type

I think that would be fine.

kytrinyx: The difference is with default fields—we would not automatically add them.

I actually think that would be fine, maybe even good. I think that it would make sense to `fmt` the whitespace and where to put braces, but not actually add fields that are implicit.

@petertseng, @NobbZ, @ErikSchierboom, @rbasso I think you all have had opinions about config formatting at one time or another. If you have a moment, would you read through this and weigh in?

robphoenix: Bypassing the Config object would mean this would only be concerned with the actual formatting rather than ensuring the JSON accurately represents the Config struct, which I think would be better. Ensures it's only doing one thing. And simpler.

kytrinyx: My concern is with the empty default fields, especially with deprecated exercises. I'd quite like to be able to have them not be in the file, as they feel more like noise than information.

robphoenix: That makes sense, changing this to @nywilken's suggestion should avoid that I think, we'll only be formatting what's in the file already.

kytrinyx: 👍 We're doing other checking for consistency in the `lint` command. I think that if we come up with more things to check we should add them there.

robphoenix:
@kytrinyx @nywilken I've refactored so as not to use the config struct, and to format the maintainers file. The downside of this is that it re-orders the JSON alphabetically. 

Example formatted Go config.json:

```json
{
  "active": true,
  "exercises": [
    {
      "core": false,
      "difficulty": 1,
      "slug": "hello-world",
      "topics": [
        "Strings"
      ],
      "unlocked_by": null,
      "uuid": "19957346-dedf-441e-85ea-656cac0d96d8"
    },
[truncated]
    {
      "deprecated": true,
      "slug": "trinary",
      "uuid": "52af9f03-8b6b-4d75-a363-124f3269dccd"
    }
  ],
  "foregone": [
    "space-age",
    "linked-list",
    "simple-linked-list"
  ],
  "ignore_pattern": "example(?!.*test)",
  "language": "Go"
}
```

Example Go maintainers.json:

```json
{
  "docs_url": "https://github.com/exercism/docs/blob/master/maintaining-a-track/maintainer-configuration.md",
  "maintainers": [
    {
      "alumnus": false,
      "avatar_url": null,
      "bio": null,
      "github_username": "bmatsuo",
      "link_text": null,
      "link_url": null,
      "name": null,
      "show_on_website": false
    },
    {
      "alumnus": true,
      "github_username": "kytrinyx"
    },
[truncated]
    {
      "alumnus": false,
      "avatar_url": null,
      "bio": "Firmware Engineer by day. I like Go a lot, and I'm optimistic about its adoption. Helping on Exercism track allows me to learn more. My focus is adding exercise solutions and test case generators.",
      "github_username": "leenipper",
      "link_text": null,
      "link_url": null,
      "name": "Lee Nipper",
      "show_on_website": true
    }
  ]
}
```

cmd/fmt.go
```diff
@@ -0,0 +1,116 @@
...
+	diff, err = difflib.GetUnifiedDiffString(
+		difflib.UnifiedDiff{
+			A: difflib.SplitLines(string(original)),
+			B: difflib.SplitLines(string(formatted)),
+		})
+	if err != nil {
+		return "", err
+	}
+
+	err = ioutil.WriteFile(path, formatted, 0644)
```
robphoenix: I don't know if this is the best way to copy the contents of the formatted JSON to the original file?

cmd/fmt.go
nywilken: I've been on a bit of a delay these past few days so just catching up. Glad to hear the suggestion was accepted and that we are all okay with some of the tradeoffs.

@robphoenix @kytrinyx I agree with the thoughts on the `lint` command doing all the consistency checks. My thoughts are that `fmt` is a helper to help track maintainers format their JSON and nothing more. Please let me know if I missed anything from the original discussion. 

Nonetheless, I think this approach would simplify things if and when we start to add more `.json` configs.

nywilken:
> @kytrinyx @nywilken I've refactored so as not to use the config struct, and to format the maintainers file. 

Nice!

> The downside of this is that it re-orders the JSON alphabetically.

I doubled checked the documentation, for maps the json output will always be sorted by the map keys - guessing this is to ensure consistency. 

> Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement encoding.TextMarshaler. The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:
> 
> - string keys are used directly
> - encoding.TextMarshalers are marshaled
> - integer keys are converted to strings

To be honest, I would be okay with this seeing as I found myself scanning the config files of different tracks to see if a key was set or not. I'm new to dealing with the config.json files and I know a lot of discussions have been had in the past about this so I will defer to you and @kytrinyx on this one.

cmd/fmt.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+func format(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
+		usageFunc(cmd)
+		return
+	}
+
+	path := args[0]
+	cfg := filepath.Join(path, "config.json")
```
nywilken: How you feel about using a slice of file paths for storing all the files? We can iterate over the slice and run format for each file? This would dry up the multiple if/else statements and make it easier for adding additional files.

cmd/fmt.go
```diff
@@ -0,0 +1,116 @@
...
+		return
+	}
+
+	maintDiff, err := formatFile(maint)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		return
+	}
+
+	if cfgDiff == "" {
```
nywilken: Do we need to inform the user if there are no changes?

My line of thinking is if there are no changes then there is nothing to display.

cmd/fmt.go
kytrinyx: I think this is fine. The file is never going to be enormous, so I think just dumping it with ioutil is good.

kytrinyx:
> The downside of this is that it re-orders the JSON alphabetically.

My first thought was "Oh, no!", but actually... if Configlet has commands to show the exercise tree and the exercise list with difficulties, I think that's fine.

cmd/fmt.go
```diff
@@ -0,0 +1,116 @@
...
+	if maintDiff == "" {
+		fmt.Printf("No differences encountered in %s.\n", maint)
+	} else {
+		fmt.Printf("Changes made in %s:\n\n%s\n", maint, maintDiff)
+	}
+	return
+}
+
+func formatFile(path string) (diff string, err error) {
+	f, err := os.Open(path)
```
nywilken: A check to see if the file exists should be added here, see [maintainer_config.go#L34](https://github.com/exercism/configlet/blob/master/track/maintainer_config.go#L34)

cmd/fmt.go
nywilken: I think overwriting the file is good. In fact, I was originally thinking to that you could use json.NewEncoder to overwrite the file. But I didn't realize the effort to get the formatted contents back. 

There is `io.Copy(dst, src)` but that would require that you have an io.Writer to the original file and an io.Reader to the formatted file.

nywilken (commented):
@robphoenix the refactor works nicely, I was able to test locally on a few tracks. Great work!

I left a few thoughts and responses to your questions. I will have time later this afternoon to do a more thorough review.

cmd/fmt.go
kytrinyx: I like this idea. Someone once said that the interesting numbers in computer science are 0, 1, 2 and infinitely many.

cmd/fmt.go
kytrinyx: Yeah, that's what I'm thinking as well. This is what gofmt, golint, and go vet all do.

kytrinyx:
> > The downside of this is that it re-orders the JSON alphabetically.

> [i]f Configlet has commands to show the exercise tree and the exercise list with difficulties, I think that's fine.

I'm going to open a discussion about this in the discussions repo, as I think it's worth making sure all the active maintainers are on board before making this call.

cmd/fmt.go
robphoenix: Ah this is a great idea @nywilken, such simplicity! :bulb: 
I like that @kytrinyx, so true.

cmd/fmt.go
robphoenix: good catch, thanks!

kytrinyx:
> I'm going to open a discussion about this

Done: https://github.com/exercism/discussions/issues/177

robphoenix:
Okay, I've updated the code in response to your suggestions @nywilken, it's a lot better for them, thanks.
I've also added a function to alphabetise each exercise's topics, I know that can be a bugbear for some.
And I've added tests to cover these changes as well.
As ever, do let me know if anything can be improved. I'll update the README soon as well. :smiley:

cmd/fmt.go
```diff
@@ -0,0 +1,131 @@
...
+		filepath.Join(path, "config", "maintainers.json"),
+	}
+
+	for _, f := range fs {
+		diff, formatted, err := formatFile(f)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err.Error())
+			return
+		}
+		if diff != "" {
```
nywilken: Nice! I like that the file is not overwritten if there are no changes.

cmd/fmt.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+func format(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
+		usageFunc(cmd)
+		return
+	}
+
+	path := args[0]
+	fs := []string{
```
nywilken: 👍

cmd/fmt.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	var m map[string]interface{}
+	json.NewDecoder(f).Decode(&m)
+
+	if filepath.Base(path) == "config.json" {
+		m = sortTopics(m)
+	}
+
+	// temporary file for encoding data into
```
nywilken: Seeing as ioutil.WriteFile is actually doing the writing the call to Encode can be dropped entirely and replaced with your original json.MarshalIndent call. More specifically, if you move lines 84 - 87 (reading the original file) to line 74 you can replace all the remaining code for creating the formatted output with a one liner. 

```Go
70   if filepath.Base(path) == "config.json" {                                                                                                                                               
71     m = sortTopics(m)                                                                                                                                                                     
72   }                                                                                                                                                                                       
73                                                                                                                                                                                           
74   original, err := ioutil.ReadFile(path)                                                                                                                                                  
75   if err != nil {                                                                                                                                                                         
76     return diff, formatted, err                                                                                                                                                           
77   }                                                                                                                                                                                       
78                                                                                                                                                                                           
79   formatted, err = json.MarshalIndent(&m, "", "  ")                                                                                                                                       
80   if err != nil {                                                                                                                                                                         
81     return diff, formatted, err                                                                                                                                                           
82   }                                                                                                                                                                                       
83                                                                                                                                                                                           
84   diff, err = difflib.GetUnifiedDiffString(
```

The caveat being that it appears to add an extra newline to the file output. Which is something we can probably just strip out.

cmd/fmt.go
```diff
@@ -0,0 +1,131 @@
...
+		for _, e := range exercises {
+			if exercise, ok := e.(map[string]interface{}); ok {
+				if topics, ok := exercise["topics"].([]interface{}); ok {
+					var sorted []string
+					for _, t := range topics {
+						if topic, ok := t.(string); ok {
+							sorted = append(sorted, topic)
+						}
+					}
+					sort.Strings(sorted)
```
nywilken: This is cool. Had to do some reading up on type assertions to make sense of this. 👍 

I tested on a few tracks and it worked nicely. The Rust track has mixed cases for some of its topics so sort doesn't work as expected. But I think that is more of config cleanup issue than a code issue. 

Nice work!

nywilken (commented):
@robphoenix your latest revision is on the money and I think it is just about ready to merge seeing as https://github.com/exercism/discussions/issues/177 is a non issue. 

I made one more suggestion to reduce the code size for the json encoding seeing as your file writing solution is more elegant than using json.Encode. 

The formatting suggestion adds an additional new line to the file outputs (which I think could be easily addressed 🤞 ). But the functionality is the same. This is purely a suggestion and is not required for merging if it adds too much work.

cmd/fmt.go
kytrinyx: I would switch this for readability to:

```
if diff == "" {
  continue
}
```

That would let you unindent the rest of it one level.

cmd/fmt.go
```diff
@@ -0,0 +1,131 @@
...
+		diff, formatted, err := formatFile(f)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err.Error())
+			return
+		}
+		if diff != "" {
+			err = ioutil.WriteFile(f, formatted, 0644)
+			if err != nil {
+				fmt.Fprintln(os.Stderr, err.Error())
+				return
```
kytrinyx: Should this be a continue as well so that we can fmt the other file if the first one fails?

kytrinyx (commented):
One more thing that I've been thinking about, is that I'd like to reduce all the JSON fixtures to the smallest possible fixture that still displays the behavior we want.

E.g. two entries for lists, an extra if we need a different feature (e.g. deprecated), short phrases instead of long sentences, "alice", "bob", and "charlie" instead of real people.

cmd/fmt.go
robphoenix: I'm not sure, should it be an all or nothing situation? I suppose though if we got to a point where we had 3 files that were being formatted, and it was the second one that failed, the first would be written, so it couldn't be an all or nothing situation, so yeah, it should.

cmd/fmt.go
robphoenix: I'd recently read about type assertions in the [gopl](https://www.amazon.co.uk/Programming-Language-Donovan-Brian-Kernighan/dp/9332569711/ref=sr_1_4?s=books&ie=UTF8&qid=1502110908&sr=1-4&keywords=go+programming+language) book, and coming from a dynamic languages background I really felt like I should be able to drill down into that data structure to access the topics. Got there with a little trial & error. I really wanted to be able to alphabetise them! 

> The Rust track has mixed cases for some of its topics so sort doesn't work as expected

I don't think there's any reason we can't normalise the topics here, save maintainers having to do it by hand or keep an eye out for topics that don't conform. What do you think @kytrinyx?

For reference:
https://github.com/exercism/discussions/issues/167
https://github.com/exercism/meta/issues/83

robphoenix:
> One more thing that I've been thinking about, is that I'd like to reduce all the JSON fixtures to the smallest possible fixture that still displays the behavior we want.

Sure, I'll redo the fixtures I've added. Is it okay to have added the files in the directory structure I have, `fixtures/format/{config,maintainers}`?

cmd/fmt.go
robphoenix: Excellent, love it when I can simplify my code, thanks!

> The caveat being that it appears to add an extra newline to the file output

Do you mean at the end of the file? I'm not seeing this behaviour. In fact it's stripping any extra newlines at the end of the file.

cmd/fmt.go
kytrinyx: > I don't think there's any reason we can't normalise the topics here, save maintainers having to do it by hand or keep an eye out for topics that don't conform.

A most excellent notion. Let's do it!

cmd/fmt.go
kytrinyx: > though if we got to a point where we had 3 files that were being formatted [...]

Yeah, that makes it even more compelling. 👍

kytrinyx:
> Is it okay to have added the files in the directory structure I have

Yeah, I like that. I've been learning that I like to keep fixtures namespaced/categorized so that we're not using one big fat set of fixtures for all the things.

cmd/fmt.go
robphoenix: It looks like `formatFile` doesn't add a newline, but a newline is added to the formatted result when the file is written.

cmd/fmt.go
robphoenix: I've added a `normaliseTopic` function. It currently lowercases, and replaces spaces with hyphens. Obviously this can be changed when a final decision has been made on how to normalise the topics in the `config.json` file.

cmd/fmt.go
```diff
@@ -0,0 +1,134 @@
...
+	f, err := os.Open(path)
+	if err != nil {
+		return diff, formatted, err
+	}
+
+	var m map[string]interface{}
+
+	json.NewDecoder(f).Decode(&m)
+
+	if format != nil {
```
robphoenix, line 90: @nywilken @kytrinyx I was uncomfortable with the `if file == config.json`, which could get potentially unwieldy if we go on to need to format specific things in other files. So I've added in the `formatter` type, so we can pass specific formatting functions to the `formatFile` function. If needs be `format` here could be a slice of `formatter`'s that it ranges over. Is this ok?

cmd/fmt.go
kytrinyx: Sounds good!

cmd/fmt.go
kytrinyx: This sounds like a good idea.

README.md
```diff
@@ -1,10 +1,12 @@
 # Configlet
 
-A linter for exercism language track repositories.
+A tool for exercism language track repositories.
```
nywilken: A tool for verifying 
exercism language track repositories.

README.md
```diff
@@ -1,10 +1,12 @@
...
 
-Exercism makes certain assumptions about language tracks. Configlet makes it simple to verify up-front that the changes to existing exercises, or the addition of new exercises will play nicely with the website.
+## Linter
 
-Configlet checks for the following configuration issues:
+Exercism makes certain assumptions about language tracks. The configlet `lint` command makes it simple to verify up-front that the changes to existing exercises, or the addition of new exercises will play nicely with the website.
```
nywilken: ... to a track's exercises, existing or newly added, will play nicely with the website.

cmd/fmt.go
nywilken: Nicely done.

cmd/fmt.go
```diff
@@ -0,0 +1,140 @@
...
+	diff, err = difflib.GetUnifiedDiffString(
+		difflib.UnifiedDiff{
+			A: difflib.SplitLines(string(original)),
+			B: difflib.SplitLines(string(formatted)),
+		})
+
+	return diff, formatted, err
+}
+
+func formatTopics(m map[string]interface{}) map[string]interface{} {
```
nywilken: A few questions here. 

1. Seeing as m is a pointer back to the original map could it be modified in place? Or because of the type assertion assignments a new map has to be returned? 

2. How would you feel about refactor to use continue statements to cut down on the level of indentations?

```Go
exercises, ok := ...
If !ok {
  return
}

for _, e := range exercises {
  exercise, ok := ...
  If !ok {
   continue 
 }
// do the same for topics, ok := ...

// Everything else could remain the same - just outdented. 

```

cmd/fmt.go
```diff
@@ -0,0 +1,140 @@
...
+					sort.Strings(sorted)
+					exercise["topics"] = sorted
+				}
+			}
+		}
+	}
+	return m
+}
+
+func normaliseTopic(t string) string {
```
nywilken: What do you think about turning this into a closure inside of the formatTopics function?

nywilken (commented):
This is coming along nicely, great work. I left a few suggestions  and a couple of questions. I still need to re-test.

README.md
kytrinyx: Maybe even _A tool for managing Exercism language track repositories_ since the scope of the tool has been expanding to be much more useful to maintainers than previously.

README.md
kytrinyx: Or maybe

... to a track's configuration, as well as changes and additions to the exercises...

nywilken: Yeah, I like that version better.

README.md
nywilken: +1

cmd/fmt.go
robphoenix: 1. Oh, why yes, yes it can. :100: 

2. I'd feel great about cutting down the indentation. :smile:

cmd/fmt.go
robphoenix: I think I'd like to keep it as a separate function. It's not set in stone yet, that this is how the topic names should be normalised, and if it changes in the future, I'd like it to be easy to update this code. I think having this logic inside a named function helps that, and hopefully helps to indicate there's a couple of things going on here, sorting the topics and normalising each one.

nywilken (commented):
@robphoenix I tested this locally on a few tracks; formatting looks great. Do you have any additional changes or refactors planned? 

If not, I think this is at a good place to squash and merge when you're ready.

nywilken (approved):

kytrinyx (approved):

cmd/fmt.go
```diff
@@ -0,0 +1,147 @@
...
+			continue
+		}
+		err = ioutil.WriteFile(f.path, formatted, os.FileMode(0644))
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err.Error())
+			continue
+		}
+		fmt.Printf("Changes made to %s:\n\n%s", f.path, diff)
+	}
+	return
```
nywilken, line 77: Sorry just a quick question, should we exit with a non zero error code if configlet is unable to find one or more of the json files? 

You can repro this by pointing configlet to some non track directory. Or a track with a missing maintainers file (I deleted the file). 

```Bash
› ./configlet fmt ~/Development/oss/exercism/rust      
stat [...]/Development/oss/exercism/rust/config/maintainers.json: no such file or directory 
› echo $?
0
```

kytrinyx: Ah, yes. That's a good point. I think we want to fail the build in that case, so a non-zero code would be important.

robphoenix: Good call, I've moved the file existence check out of the `formatFile` function and added a `os.Exit(1)`

robphoenix:
@kytrinyx @nywilken Thankyou both for your patience and insightful comments, this has been a great experience, I've learnt a lot from it. :heart:

kytrinyx:
> this has been a great experience, I've learnt a lot from it.

Me too! This has been so great.

nywilken:
Ditto - an awesome experience for sure.

PR #36, opened 2017-07-29
Delete unused variable

kytrinyx:
The error message needs dynamic data (path, actual error), so this variable never ended up
getting used.

### DIFF
--- a/track/config.go
+++ b/track/config.go
@@ -2,12 +2,9 @@ package track
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 )
-
-var errInvalidConfig = errors.New("invalid config file - try jsonlint.com")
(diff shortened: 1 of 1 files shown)


nywilken:
Question about Go and not the code. Normally, Go will throw a compile error for unused variables but that doesn't seem to be true for package level vars. Do you know the rationale?

nywilken (approved):
This looks good. I left a question about Go in general.

kytrinyx:
> Do you know the rationale?

No, I'm not sure. I was thinking that it might be because other packages that use this one as a library might want to set values, but that doesn't hold for unexported variables.

nywilken:
Found the answer https://github.com/golang/go/issues/222

kytrinyx:
Nice!

PR #45, opened 2017-08-07
Minor changes to lint & generate commands

robphoenix:
* use filepath.FromSlash in lint_example_test
* Add usage functions to lint & generate commands


kytrinyx (approved):
Good call. Thanks for cleaning that up!

PR #48, opened 2017-08-11
Provide a MixedCase version of the problem spec name

kytrinyx:
This will let us use that in generated READMEs for languages who name their test files in mixed case.

E.g. here: https://github.com/exercism/groovy/tree/master/exercises/difference-of-squares


nywilken (approved):
This looks good. I tested readme generation on the Groovy track, with the instructions provided in exercism/groovy#63, to confirm the fix. Merging 👍

kytrinyx:
Awesome. I'm going to cut a release in a moment so that we can go make the fix to the Groovy track. Thanks @nywilken!

PR #49, opened 2017-08-15
Add check for missing problem-specifications directory

### LINKED ISSUE #27: Print error message if problem-specifications is not found
Present README.md looks like [this](https://github.com/exercism/delphi/blob/master/exercises/wordy/README.md).

New README.md looks like [this](https://github.com/rpottsoh/exercism-delphi/blob/configletWordyREADME/exercises/wordy/README.md).

Should all the information regarding the problem be removed from the resulting README.md file?  Is there something that I should change in [exercise_readme.go.tmpl](https://github.com/exercism/delphi/blob/master/config/exercise_readme.go.tmpl)?

nywilken:
Readme generation depends on the existence of the problem-specifications repository. This change fixes an issue where executing a readme generation with an zero-value ProblemSpecification results in a incomplete Readme.md file. 

The configlet command will now print an error message if it is unable to load the problem-specifications repository.

Closes exercism/configlet#27

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 
 	"github.com/exercism/configlet/track"
+	"github.com/hashicorp/go-multierror"
 	"github.com/spf13/cobra"
 )
 
@@ -35,13 +36,20 @@ func generate(cmd *cobra.Command, args []string) {
 		fmt.Fprintf(os.Stderr, err.Error())
 		os.Exit(1)
 	}
-
-	track.ProblemSpecificationsPath = specPath
 	root := filepath.Dir(path)
 	trackID := filepath.Base(path)
 
-	var exercises []track.Exercise
+	track.ProblemSpecificationsPath = filepath.Join(root, track.ProblemSpecificationsDir)
+	if specPath != "" {
+		track.ProblemSpecificationsPath = specPath
+	}
 
+	if _, err := os.Stat(track.ProblemSpecificationsPath); os.IsNotExist(err) {
+		fmt.Fprintf(os.Stderr, "path not found: %s\n", track.ProblemSpecificationsPath)
+		os.Exit(1)
+	}
+
+	var exercises []track.Exercise
 	if genSlug != "" {
 		exercises = append(exercises, track.Exercise{Slug: genSlug})
 	} else {
@@ -53,17 +61,24 @@ func generate(cmd *cobra.Command, args []string) {
 		exercises = track.Exercises
 	}
 
-	errs := []error{}
+	errs := &multierror.Error{}
 	for _, exercise := range exercises {
 		readme, err := track.NewExerciseReadme(root, trackID, exercise.Slug)
 		if err != nil {
-			errs = append(errs, err)
+			errs = multierror.Append(errs, err)
 			continue
 		}
+
 		if err := readme.Write(); err != nil {
-			errs = append(errs, err)
+			errs = multierror.Append(errs, err)
 		}
 	}
+
+	if err := errs.ErrorOrNil(); err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err.Error())
+		os.Exit(1)
+	}
+
 }
 
 func init() {
--- a/track/exercise_readme.go
+++ b/track/exercise_readme.go
@@ -40,15 +40,15 @@ func NewExerciseReadme(root, trackID, slug string) (ExerciseReadme, error) {
 		dir:      filepath.Join(root, trackID, dirExercises, slug),
 	}
 
+	spec, err := NewProblemSpecification(root, trackID, slug)
+	if err != nil {
+		return readme, err
+	}
+	readme.Spec = spec
+
 	if err := readme.readTemplate(); err != nil {
 		return readme, err
 	}
-
-	spec, err := NewProblemSpecification(root, trackID, slug)
-	if err != nil {
-		spec = &ProblemSpecification{}
-	}
-	readme.Spec = spec
 
 	if err := readme.readHints(); err != nil {
 		return readme, err
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -8,6 +8,18 @@
   version = "v1.1.0"
 
 [[projects]]
+  branch = "master"
+  name = "github.com/hashicorp/errwrap"
+  packages = ["."]
+  revision = "7554cd9344cec97297fa6649b055a8c98c2a1e55"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/hashicorp/go-multierror"
+  packages = ["."]
+  revision = "83588e72410abfbe4df460eeb6f30841ae47d4c4"
+
+[[projects]]
   name = "github.com/inconshreveable/mousetrap"
   packages = ["."]
   revision = "76626ae9c91c4f2a10f34cad8ce83ea42c93bb75"
@@ -52,6 +64,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "9c9b4be2565453cbd659554e7905e2a9a413f5d1de43914204330564f9cbc093"
+  inputs-digest = "86718b7614f4254ac2e62b1cd50e8886d1b062122f49da8fca8a0a76aef80884"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/fixtures/problem-specifications/exercises/fake/description.md
+++ b/fixtures/problem-specifications/exercises/fake/description.md
@@ -1,0 +1,1 @@
+This is fake.
--- a/fixtures/problem-specifications/exercises/fake/metadata.yml
+++ b/fixtures/problem-specifications/exercises/fake/metadata.yml
@@ -1,0 +1,4 @@
+---
+blurb: "Fake."
+source: "The internet."
+source_url: "http://example.com"
--- a/track/exercise_readme_test.go
+++ b/track/exercise_readme_test.go
@@ -51,6 +51,7 @@ func TestExerciseReadmeTrackInsertDeprecation(t *testing.T) {
 		{"inserts-old", "deprecated insert\n"},
 	}
 
+	ProblemSpecificationsPath = filepath.FromSlash("../fixtures/problem-specifications")
 	for _, test := range tests {
 		readme, err := NewExerciseReadme(root, test.trackID, "fake")
 		assert.NoError(t, err)
@@ -69,6 +70,7 @@ func TestExerciseReadmeHintsDeprecation(t *testing.T) {
 		{"hints-old", "deprecated hints\n"},
 	}
 
+	ProblemSpecificationsPath = filepath.FromSlash("../fixtures/problem-specifications")
 	for _, test := range tests {
 		readme, err := NewExerciseReadme(root, test.trackID, "fake")
 		assert.NoError(t, err)
--- a/track/problem_specification.go
+++ b/track/problem_specification.go
@@ -10,8 +10,10 @@ import (
 )
 
 const (
-	filenameDescription = "description.md"
-	filenameMetadata    = "metadata.yml"
+	// ProblemSpecificationsDir is the default name of the cloned problem-specifications repository.
+	ProblemSpecificationsDir = "problem-specifications"
+	filenameDescription      = "description.md"
+	filenameMetadata         = "metadata.yml"
 )
 
 var (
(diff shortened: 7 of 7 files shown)


nywilken:
@kytrinyx @robphoenix I saw this issue and took the opportunity to divide into how readme generation works a little further. I plan to add a test or two to verify error handling but, as always, I'm open to initial thoughts and feedback.

cmd/generate.go
```diff
@@ -62,8 +62,12 @@ func generate(cmd *cobra.Command, args []string) {
 			errs = append(errs, err)
 			continue
 		}
+
+		// TODO find the best way to handle multiple errors.
```
kytrinyx: There's a multi errors library from hashicorp that I've used successfully before.

track/exercise_readme.go
```diff
@@ -40,16 +40,15 @@ func NewExerciseReadme(root, trackID, slug string) (ExerciseReadme, error) {
 		dir:      filepath.Join(root, trackID, dirExercises, slug),
 	}
 
+	// Don't care if this errors here as errors are handled, where it counts, in readme.Generate().
+	if err := readme.loadProblemSpecification(); err != nil {
```
kytrinyx: Do we load the problem specification twice, once here and once in Generate?

track/exercise_readme.go
```diff
@@ -63,6 +62,10 @@ func NewExerciseReadme(root, trackID, slug string) (ExerciseReadme, error) {
 
 // Generate produces a README from the template and data.
 func (readme ExerciseReadme) Generate() (string, error) {
+	if err := readme.loadProblemSpecification(); err != nil {
+		return "", err
+	}
+
```
kytrinyx: This is loading it specifically so that it can check the error, right?

Instead of doing that, maybe we can have a specific test at the command level that says "problem specification directory exists".

I guess that would make it possible to have an exercise that is missing metadata entirely, but that feels like a different problem (linting, perhaps).

track/exercise_readme.go
```diff
@@ -82,6 +85,26 @@ func (readme ExerciseReadme) Write() error {
 	return ioutil.WriteFile(filepath.Join(readme.dir, filenameReadme), []byte(s), 0644)
 }
 
+// loadProblemSpecification locates Problem Specification repository data and initializes Spec for readme.
+// If Spec for readme has already been initialized loadProblemSpecification does nothing.
+func (readme *ExerciseReadme) loadProblemSpecification() error {
```
kytrinyx: Oh, I see. Right. Hm.

I think I still would like to be explicit about checking for the existence of the problem-specifications repository up front rather than burying this down here with extra caching.

cmd/generate.go
nywilken: Sweet, thanks! Guessing you're referring to https://github.com/hashicorp/go-multierror

track/exercise_readme.go
nywilken: So problem specification is only loaded once, but the call is made twice to support the loading of the problem specification upon ExerciseReadme construction. I had a version of the code without loading problem specification in the constructor - only within readme.Generate() -  but added it back once I saw that readme.Spec would remain nil until calling readme.Generate(). 

Was the original intent to not have ProblemSpecification be a hard dependency for the initialization of an ExerciseReadme? In other words, it should not error if the the loading of the problem specification failed? 

I ask because, I thought to error here but the original code sort of swallowed the error and set it to an empty ProblemSpecification. What do you think about making it a hard dependency, if problem specification fails to load return immediately like we do for readTemplate()?

track/exercise_readme.go
nywilken: > This is loading it specifically so that it can check the error, right?

Yes.

> Instead of doing that, maybe we can have a specific test at the command level that says "problem specification directory exists".

I agree with this approach. I left a comment, above, about my first implementation that I would like to get your thoughts on.

track/exercise_readme.go
nywilken: I agree, explicit is the way to go. Thanks for reviewing.

track/exercise_readme.go
kytrinyx: I'm fine with making it a hard dependency. I think that making it a not-hard dependency was accidentaly.

kytrinyx: Oh, also: this would mean that we don't need a specific "can't find the problem spec dir" test, though that might be easier to debug than an error coming from farther down. I don't know though—if the error message is "can't find a problem spec" then it might well be just as easy to see what's up.

cmd/generate.go
kytrinyx: Yeah! That's the one. Thanks for finding it—I've had a very confusing day and got distracted before I remembered to do that.

track/exercise_readme.go
nywilken: Great, thanks for your thoughts here. I have a working version that I hope to push up either tonight or in the am - been mobile pretty much the whole day. 

As for where the error message should live, I think in both places is probably the best for correctness. 

In the generate command, if spec-path doesn't exist we should error with a friendly message "hmmm, seems like the problem specification path provided doesn't exist; check the path and try again". This would exit immediately as there is no point in trying to initialize an exercise readme if we know it will fail. 

In the case where spec-path does exist, the check within the the exercise readme will error if it can't read the problem specification data for one, more or all exercises. Preventing the generation of an incomplete readme. 

Sounds good to you?

kytrinyx: Yepp, that sounds good to me!

cmd/generate.go
```diff
@@ -32,11 +33,20 @@ var (
 func generate(cmd *cobra.Command, args []string) {
 	path, err := filepath.Abs(filepath.FromSlash(args[0]))
 	if err != nil {
-		fmt.Fprintf(os.Stderr, err.Error())
+		fmt.Println(err.Error())
```
kytrinyx: Is it intentional to put this onto stdout instead of stderr?

(Side note, I'm pretty sure you can say `fmt.Println(err)` rather than `fmt.Println(err.Error())`)

cmd/generate.go
```diff
@@ -32,11 +33,20 @@ var (
...
 		os.Exit(1)
 	}
 
-	track.ProblemSpecificationsPath = specPath
+	if specPath != "" {
+		_, err := os.Stat(specPath)
+		if os.IsNotExist(err) {
+			fmt.Printf("hmm, it looks like %s doesn't exist; please check the path and try again.", specPath)
```
kytrinyx: I'm wondering if it's worth shortening this a bit, especially since the path is likely to be quite long. Maybe just:

> Path %s not found.

I suppose best practice says that we should suggest how to fix the error, but in this case I think that "not found" kind of obviously means "check that it's actually there".

cmd/generate.go
```diff
@@ -32,11 +33,20 @@ var (
 func generate(cmd *cobra.Command, args []string) {
 	path, err := filepath.Abs(filepath.FromSlash(args[0]))
 	if err != nil {
-		fmt.Fprintf(os.Stderr, err.Error())
+		fmt.Println(err.Error())
 		os.Exit(1)
 	}
 
-	track.ProblemSpecificationsPath = specPath
+	if specPath != "" {
```
kytrinyx: Thinking out loud, let's set the specPath if it's not empty, then check the existence of `track.ProblemSpecificationsPath`, afterwards.

This will catch the problem in both cases (whether they've specified a custom location, or just using the default location).

kytrinyx (commented):
I think this looks good. I have a couple of questions/thoughts.

cmd/generate.go
nywilken: > Is it intentional to put this onto stdout instead of stderr?

Yeah, I realized when testing the multi error output that I couldn't pipe the error messages to less without redirecting `... 2>&1` so I  changed to use Stodout. 

But maybe the correct thing to do here is implement a function for displaying multiple errors with a different level of verbosity. Or add a verbosity flag. I'm open to suggestions. 

> Side note ...

Good catch, I moved from Fprintf but didn't change the argument.

cmd/generate.go
nywilken: ProblemSpecificationsPath is an empty string by default and ProblemSpecification will only use it if it is set to anything other than empty string. 

Which makes sense seeing as it at least needs to know some root path, which is not known until initializing. 

Let me think about this a little to see if we can refactor and if it makes sense to.

cmd/generate.go
nywilken: > so I changed to use Stodout.

In hindsight, this should probably be it's own PR so we can discuss Stderr vs Stdout for errors. I will break out this commit into its own PR.

cmd/generate.go
nywilken: I'm with you on making this shorter 🙌🏽
I had to resize my tmux pane a few times just to read 😂  

`Path Not Found` is perfect.

cmd/generate.go
kytrinyx: Thanks, yes, I'd like to discuss this 🌻

cmd/generate.go
kytrinyx: > ProblemSpecificationsPath is an empty string by default and ProblemSpecification will only use it if it is set to anything other than empty string.

How about we have an `init` in the `track` package that sets the ProblemSpecPath to the default?

cmd/generate.go
kytrinyx: > Path Not Found

The error should not be title cased, so `path not found`. Shouldn't be sentence cased either, so I was wrong in my comment. From the official code review comments: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

cmd/generate.go
nywilken: That would be ideally, but we need to know at least the root path to the track, before calling init which wouldn't be possible without running the command with some track path, right? 

So instead I added a constant `ProblemSpecificationsDir` to the track package that can be used in the generate command to build the default ProblemSpecificationsPath based off the provided track path. 

```
track.ProblemSpecificationsPath = filepath.Join(root, track.ProblemSpecificationsDir)
// check if track.ProblemSpecificationsPath exists; if no error out 
...
```

I will push up for testing and feedback.

nywilken: Oh a quick question, was there some plans on storing the ProblemSpecPath in the ProblemSpecification type? I see https://github.com/exercism/configlet/blob/master/track/problem_specification.go#L32 but it is not used nor set.

cmd/generate.go
nywilken: ~~Sorry to nit pick here but seeing as we are displaying an error to the user, similar to the log call in the example, and not returning it to some other caller I think it may be okay to format the error message a little differently. What do you think about using something similar to what we do in the lint command  `-> The path '/for/bar' does not exist.`?~~

Sticking with what we got; it works.

nywilken:
@kytrinyx @robphoenix I reverted the changes for printing errors and changed the logic a bit for ProblemSpecificationsPath. Let me know what you think, thanks.

cmd/generate.go
kytrinyx: Yeah, we should probably have a discussion at some point about what we do with errors when they are the last stop and we know we're going to show them to the user.

kytrinyx (approved):
This looks great to me.

🚢

cmd/generate.go
```diff
@@ -35,13 +36,20 @@ func generate(cmd *cobra.Command, args []string) {
...
 	trackID := filepath.Base(path)
 
-	var exercises []track.Exercise
+	track.ProblemSpecificationsPath = filepath.Join(root, track.ProblemSpecificationsDir)
+	if specPath != "" {
+		track.ProblemSpecificationsPath = specPath
+	}
 
+	if _, err := os.Stat(track.ProblemSpecificationsPath); os.IsNotExist(err) {
+		fmt.Fprintf(os.Stderr, "path '%s' not found\n", track.ProblemSpecificationsPath)
```
robphoenix: a little nitpick; with this error message I think the reason should be separate from the cause, ie. `path not found: %s` rather than the path within the error message. This seems to be more idiomatic, and also is easier to follow I think, as well as if the path is long then the `not found` part could be harder to notice.

Though I notice @kytrinyx mention a future discussion on best practice for user consumed error messages, so this can wait til then if you like.

robphoenix (approved):
Looks great, one tiny little thing, otherwise I concur with @kytrinyx :ship:

cmd/generate.go
kytrinyx: I think `path not found: %s` is an excellent call.

We should still have that discussion about user-consumed messages eventually, though!

nywilken: This is a simple enough change so I will update to `path not found: %s`, before merging. Thanks 🙏 

FYI, I spent a little time reading up on the Stderr vs Stdout and came across this great post https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

After reading, I think deferring to Doug McIlory is the way to go. So I'm going to rescind my Stdout PR, before even creating it.

kytrinyx: > After reading, I think deferring to Doug McIlory is the way to go. So I'm going to rescind my Stdout PR, before even creating it.

❤️

kytrinyx: > I will update to path not found: %s, before merging. 

Perfect. Merge when you're happy, I'd say.

PR #52, opened 2017-08-17
Add verbose flag to fmt

robphoenix:
The diff output can sometimes be quiet noisy and may not always be
desired. This changes it so that this noisyness only happens when
explicitly requested with the `verbose` flag.


kytrinyx (approved):
This is a great idea!

nywilken (approved):

PR #53, opened 2017-08-17
Add verbose flag to fmt.

robphoenix:
With this the default output if there are changes will only be
the filenames where changes have occured. To include the actual
diff in the output, the `verbose` flag can be used.


robphoenix:
Argh, `hub pull-request` messing with me, sorry.

PR #54, opened 2017-08-17
Fix generate usage output

robphoenix:
Using a usage function was overriding the default output which included the commands flags. 
This PR removes the usage function in favour of defining the usage in the initial command struct.

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -17,21 +17,19 @@ var (
 var (
 	// generateCmd represents the generate command
 	generateCmd = &cobra.Command{
-		Use:   "generate",
+		Use:   "generate <path/to/track>",
 		Short: "Generate exercise READMEs for an Exercism language track",
-		Long: `Generate READMEs for Exercism exercises based on contents of
-a number of different files.`,
-		Run: generate,
+		Long:  `Generate READMEs for Exercism exercises based on the contents of various files.`,
+		Example: `  configlet generate <path/to/track> --only hello-world
+
+  configlet generate <path/to/track> --spec-path <path/to/problem-specifications>
+`,
+		Run:  generate,
+		Args: cobra.MinimumNArgs(1),
 	}
-	generateUsageText = "USAGE: configlet generate <path/to/track>\n"
 )
 
 func generate(cmd *cobra.Command, args []string) {
-	if len(args) == 0 {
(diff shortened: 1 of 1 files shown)


cmd/generate.go
```diff
@@ -17,21 +17,19 @@ var (
...
-		Long: `Generate READMEs for Exercism exercises based on contents of
-a number of different files.`,
-		Run: generate,
+		Long:  `Generate READMEs for Exercism exercises based on the contents of various files.`,
+		Example: `  configlet generate <path/to/track> --only hello-world
+
+  configlet generate <path/to/track> --spec-path <path/to/problem-specifications>
+`,
+		Run:  generate,
+		Args: cobra.MinimumNArgs(1),
```
robphoenix, line 28: This ensures the command will error if there is less than 1 arg.

robphoenix:
Until we choose to use a UsageTemplate then I believe this is the best way to go. I'll update the other commands if this is successfully approved & merged.

This is useful reference for https://github.com/exercism/cli/issues/442 also.

robphoenix:
Updated help text for reference:

```
Generate READMEs for Exercism exercises based on the contents of various files.

Usage:
  configlet generate <path/to/track> [flags]

Examples:
  configlet generate <path/to/track> --only hello-world

  configlet generate <path/to/track> --spec-path <path/to/problem-specifications>


Flags:
  -h, --help               help for generate
  -o, --only string        Generate READMEs for just the exercise specified (by the slug).
  -p, --spec-path string   The location of the problem-specifications directory.
```

cmd/generate.go
nywilken: Guessing this is to ensure that at least <path/to/track> is provided?

nywilken (approved):
@robphoenix thanks for pushing this through. I saw this issue while working on the problem specification bug. TBH, we should probably drop the usageFunc from other commands until, like you said, pick a UsageTemplate otherwise we loose the default functionality. ~~Can you merge this in if you're ready as I would like to rebase on my branch before I submit~~. I'll merge this in now. 

FYI, I will be introducing the http://godoc.org/github.com/hashicorp/go-multierror pkg for displaying multiple errors in my PR.

cmd/generate.go
robphoenix: yep

PR #55, opened 2017-08-17
Refactor usage output for all commands

robphoenix:
See https://github.com/exercism/configlet/pull/54

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -18,8 +18,8 @@ var fmtVerbose bool
 
 // fmtCmd defines the fmt command
 var fmtCmd = &cobra.Command{
-	Use:   "fmt",
-	Short: "Format the track configuration files.",
+	Use:   "fmt " + pathExample,
+	Short: "Format the track configuration files",
 	Long: `The fmt command formats the track's configuration files.
 
 It ensures the following files have consistent JSON syntax and indentation:
@@ -27,21 +27,15 @@ It ensures the following files have consistent JSON syntax and indentation:
 	
 It also normalizes and alphabetizes the exercise topics in the config.json file.
 `,
-	Run: format,
+	Example: fmt.Sprintf("  %s fmt %s --verbose", binaryName, pathExample),
+	Run:     runFmt,
+	Args:    cobra.MinimumNArgs(1),
 }
 
 // formatter applies additional formatting to unmarshalled JSON files.
 type formatter func(m map[string]interface{})
 
-// formatUsageText defines how to use the fmt command
-var formatUsageText = "Usage:\n  configlet fmt <path/to/track>\n"
-
-func format(cmd *cobra.Command, args []string) {
-	if len(args) == 0 {
-		formatUsageFunc(cmd)
-		return
-	}
-
+func runFmt(cmd *cobra.Command, args []string) {
 	path := args[0]
 	var fs = []struct {
 		path string
@@ -155,13 +149,7 @@ func normaliseTopic(t string) string {
 	return s
 }
 
-func formatUsageFunc(cmd *cobra.Command) error {
-	fmt.Fprintf(os.Stderr, formatUsageText)
-	return nil
-}
-
 func init() {
 	RootCmd.AddCommand(fmtCmd)
-	fmtCmd.SetUsageFunc(formatUsageFunc)
 	fmtCmd.Flags().BoolVarP(&fmtVerbose, "verbose", "v", false, "display the diff of the formatted changes.")
 }
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -10,8 +10,8 @@ import (
 
 // lintCmd defines the lint command.
 var lintCmd = &cobra.Command{
-	Use:   "lint",
-	Short: "Ensure that the track is configured.",
+	Use:   "lint " + pathExample,
+	Short: "Ensure that the track is configured correctly",
 	Long: `The lint command checks for any discrepancies in a track's configuration files.
 
 It ensures the following files are valid JSON:
@@ -19,17 +19,12 @@ It ensures the following files are valid JSON:
 
 It also checks that the exercises defined in the config.json file are complete.
 `,
-	Run: lint,
+	Example: fmt.Sprintf("  %s lint %s", binaryName, pathExample),
+	Run:     runLint,
+	Args:    cobra.MinimumNArgs(1),
 }
 
-// lintUsageText defines how to use the lint command
-var lintUsageText = "Usage:\n  configlet lint <path/to/track>\n"
-
-func lint(cmd *cobra.Command, args []string) {
-	if len(args) == 0 {
-		lintUsageFunc(cmd)
-		return
-	}
+func runLint(cmd *cobra.Command, args []string) {
 	var hasErrors bool
 	for _, arg := range args {
 		if failed := lintTrack(arg); failed {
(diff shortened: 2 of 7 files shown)


cmd/fmt.go
```diff
@@ -27,21 +27,15 @@ It ensures the following files have consistent JSON syntax and indentation:
 	
 It also normalizes and alphabetizes the exercise topics in the config.json file.
 `,
-	Run: format,
+	Example: `  configlet fmt ~/exercism/track --verbose`,
+	Run:     format,
+	Args:    cobra.MinimumNArgs(1),
```
kytrinyx: I had no idea! This is great.

cmd/fmt.go
```diff
@@ -27,21 +27,15 @@ It ensures the following files have consistent JSON syntax and indentation:
 	
 It also normalizes and alphabetizes the exercise topics in the config.json file.
 `,
-	Run: format,
+	Example: `  configlet fmt ~/exercism/track --verbose`,
```
kytrinyx: Should we use `os.Args[0]` for the binary name in case people are naming it differently?

cmd/lint.go
```diff
@@ -19,17 +19,12 @@ It ensures the following files are valid JSON:
 
 It also checks that the exercises defined in the config.json file are complete.
 `,
-	Run: lint,
+	Example: `  configlet lint ~/exercism/track`,
```
kytrinyx: I like having the example. We should definitely do that with all of them (and for the CLI).

cmd/root.go
```diff
@@ -9,9 +9,14 @@ import (
...
-	Short: "A configuration tool for Exercism tracks.",
-	Long:  `Configlet helps verify that Exercism tracks are configured correctly.`,
+	Use:   "configlet [command]",
+	Short: "A tool for managing Exercism language track repositories.",
+	Long: `Configlet helps verify that Exercism tracks are configured correctly.
+
+The lint command verifies changes to a track's configuration, and ensures
+changes and additions to exercises will play nicely with the website.  The fmt
+command consistently formats a track's configuration files.
+`,
```
kytrinyx: Should we mention generate here?

cmd/uuid.go
```diff
@@ -13,15 +13,19 @@ var uuidCmd = &cobra.Command{
...
-	`,
-	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println(uuid.GenUUID())
-	},
+Each Exercism exercise needs a unique and unmutable UUID. This needs to be
+unique across the entire platform, not just within a track. In other words, if
+you have 'clock' in Go and 'clock' in Haskell, they need to have different
+UUIDs, even though they are based on the same problem specification.
+`,
+	Run:  UUID,
```
kytrinyx: The other commands have unexported functions that they're using for the Run function. Should we make this `uuid` instead of `UUID`?

kytrinyx (commented):
This is a great change. I have a few thoughts, nothing revolutionary or particularly insightsful :)

cmd/lint.go
nywilken: As we move through the messaging and usage functions, examples should be provided where possible. They really help jump start a user who may be new to the configlet / exercism cli.

cmd/root.go
nywilken: Do you think the long description would be repetitive since calling configlet with with no command displays the help text, which contains Available Commands?

```
› ./configlet
Configlet helps verify that Exercism tracks are configured correctly.

Usage:
  configlet [command]

Available Commands:
  fmt         Format the track configuration files.
  generate    Generate exercise READMEs for an Exercism language track
  help        Help about any command
  lint        Ensure that the track is configured.
  uuid        Generate a UUID.
  version     Output the current version of the tool.

Flags:
  -h, --help   help for configlet

Use "configlet [command] --help" for more information about a command.
```

cmd/uuid.go
nywilken: I would vote for consistency here. We should also rename  the function as the command is already using the uuid package to call uuid.GenUUID().

cmd/root.go
kytrinyx: > Do you think the long description would be repetitive since calling configlet with with no command displays the help text, which contains Available Commands?

Good point. Yeah, probably redundant.

cmd/uuid.go
kytrinyx: Dang. I liked having the function be named for the command, but that won't work here.

So, maybe `runUUID`, `runLint`, `runFmt`, and `runGenerate`?

nywilken: I'm cool with that.

cmd/fmt.go
robphoenix: If we do, it jars with the `Use:` field as it will use `configlet` so we get 
```
Usage:
  configlet fmt <path/to/track> [flags]

Examples:
  ./bruce fmt ~/exercism/track --verbose
```
I think this is something that can be remedied if we start using a custom `UsageTemplate` but I've not had time to look into that properly yet. I could be missing something with regards to where Cobra is pulling in the binary name, otherwise for now those renaming it should hopefully understand that the tool is called configlet whatever they've renamed it to.

cmd/uuid.go
robphoenix: Me too.

cmd/root.go
robphoenix: Agree. Current help output:
```
Configlet v3.3.0

A tool for managing Exercism language track repositories.

Usage:
  configlet [command]

Examples:
  configlet fmt <path/to/track>  
	
  configlet generate <path/to/track>
  
  configlet lint <path/to/track>

Available Commands:
  fmt         Format the track configuration files
  generate    Generate exercise READMEs for an Exercism language track
  help        Help about any command
  lint        Ensure that the track is configured correctly
  uuid        Generate a UUID
  version     Output the current version of the tool

Flags:
  -h, --help   help for configlet

Use "configlet [command] --help" for more information about a command.
```
I've included the `Version` variable.

robphoenix: I didn't include `uuid` & ` version` in the examples as the list was starting to get long and they're pretty self-explanatory.

cmd/fmt.go
kytrinyx: Cobra pulls the binary name from `Use`, which we can override:

```
// RootCmd is the main deal.
var RootCmd = &cobra.Command{
	Use:   "echeques",
	Short: "A brief description of your application",
	Long:  "A brief description of your application",
	Run:   func(cmd *cobra.Command, args []string) { fmt.Println("called the command") },
}

func init() {
	if len(os.Args) > 0 {
		RootCmd.Use = os.Args[0]
	}
}

```

```
go build -o chess main.go 
```

```
./chess -h
```

```
A brief description of your application

Usage:
  ./chess [flags]

Flags:
  -h, --help   help for ./chess
```

robphoenix: For some reason I couldn't seem to get this `init` function to work in the same way you have here. So I've created a var (`binName`) that holds `os.Args[0]` and anywhere `configlet` was previously used in the help/usage text has now been replaced by `binName`. This makes sure there's no occasion where both `configlet` and a user given bin name appear in the help/usage text.

cmd/root.go
```diff
@@ -7,16 +7,18 @@ import (
...
+
 // RootCmd represents the base command when called without any subcommands
 var RootCmd = &cobra.Command{
-	Use:   "configlet [command]",
+	Use:   binName,
 	Short: "A tool for managing Exercism language track repositories.",
-	Long:  fmt.Sprintf("Configlet v%s\n\nA tool for managing Exercism language track repositories.", Version),
-	Example: `  configlet fmt <path/to/track>  
+	Long:  fmt.Sprintf("%s v%s\n\nA tool for managing Exercism language track repositories.", binName, Version),
+	Example: fmt.Sprintf(`  %s fmt <path/to/track>
```
kytrinyx: If find it a bit tricky to read when the `fmt.Sprintf()` contains a multiline string. Should we extract the example into a function?

robphoenix: Yep, that is cleaner.

kytrinyx (approved):
This looks great!

Issue #30, opened 2017-07-23
Update lint command: verify UUIDs against all other tracks


kytrinyx:
The UUIDs in a track must not collide with UUIDs in other tracks. In other words, the UUID for `clock` in Go must be distinct from the UUID for `clock` in Haskell. Or any other exercise, for that matter.

I've added a UUID verification service to the Exercism API

The `lint` command should call the UUID service with the expected payload, where `uuids` is an array containing all the exercise UUIDs in the track.

The command should accept a flag `--no-http` which will skip this verification (in case people are working without internet).

We might also consider providing a flag `--uuid-service` which would call a different URL (presumably something on localhost). We could very easily provide a small service that could run locally.

#### UUID Service

```
curl -v -XPOST  "http://exercism.io/api/v1/uuids" -H "Content-Type: application/json" --data '{                              
"track_id": "ruby",
"uuids": ["af961c87-341c-4dd3-a1eb-272501b9b0e4"]
}'
...
< HTTP/1.1 204 No Content
...
```

The service responds with HTTP status 204 if all is well, or 409 if there is a conflict. In the case of a conflict the payload contains the conflicting UUIDs.

```
 curl -v -XPOST  "http://exercism.io/api/v1/uuids" -H "Content-Type: application/json" --data '{                              
"track_id": "go",
"uuids": [
  "bf2dcbc4-266f-4cad-9cd7-76d446360a5d",
  "af961c87-341c-4dd3-a1eb-272501b9b0e4",
  "9a06556b-a661-4329-97a1-b5d0f0c7590c"
  ]
}'
...
< HTTP/1.1 409 Conflict
...
{"uuids":["af961c87-341c-4dd3-a1eb-272501b9b0e4"]}
```

PR #57, opened 2017-08-18
Add UUID validation across exercises and tracks to lint command

### LINKED ISSUE #28: Update lint command: verify that each exercise has a UUID
When verifying the `config.json` for a track, each exercise is expected to have a UUID.

If it does not, the command should output an error message that explains that exercise `$SLUG` does not have a UUID.

### LINKED ISSUE #29: Update lint command: reject duplicate exercise UUIDs
Copy/paste is probably the most common way of adding a new `exercise` in a track's `config.json` file. This means that we're in great danger of copy/pasted (and therefore duplicated) UUIDs.

If the `lint` command detects that more than one exercise have the same UUID, it should complain, mentioning which exercises.

It could also generate new UUID suggestions to make it easy to fix (one suggestion per duplicate).

There is a new `configlet uuid` command that generates UUIDs, take a look at `cmd/uuid.go` to see the function call to use for suggestions.

nywilken:
This is a simple check that validates exercises UUIDs against the metadata defined within the track's config.json.

- [x] Fix fixtures now that UUID is required. 
- [x] Review logic for API to check UUIDs across the wire.

closes #28, #29, #30

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -1,12 +1,21 @@
 package cmd
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 
 	"github.com/exercism/configlet/track"
 	"github.com/spf13/cobra"
 )
+
+// UUIDValidationURL is the endpoint to Exercism's UUID validation service.
+var UUIDValidationURL = "http://exercism.io/api/v1/uuids"
+
+// disableHTTPChecks flag skips HTTP based checks when passed.
+var disableHTTPChecks bool
 
 // lintCmd defines the lint command.
 var lintCmd = &cobra.Command{
@@ -78,16 +87,24 @@ func lintTrack(path string) bool {
 			check: duplicateSlugs,
 			msg:   "-> The exercise '%v' was found in multiple (conflicting) categories in config.json.\n",
 		},
+		{
+			check: duplicateUUID,
+			msg:   "-> The following UUID occurs multiple times. Each exercise UUID must be unique.\n%v\n",
+		},
+		{
+			check: duplicateTrackUUID,
+			msg:   "-> The following UUID was found in multiple Exercism tracks. Each exercise UUID must be unique across tracks.\n%v\n",
+		},
 	}
 
 	hasErrors := false
 	for _, configError := range configErrors {
-		slugs := configError.check(t)
+		failedItems := configError.check(t)
 
-		if len(slugs) > 0 {
+		if len(failedItems) > 0 {
 			hasErrors = true
-			for _, slug := range slugs {
-				fmt.Printf(configError.msg, slug)
+			for _, item := range failedItems {
+				fmt.Printf(configError.msg, item)
 			}
 		}
 	}
@@ -228,6 +245,73 @@ func duplicateSlugs(t track.Track) []string {
 	return slugs
 }
 
+func duplicateUUID(t track.Track) []string {
+	uuids := []string{}
+	seen := map[string]bool{}
+	for _, exercise := range t.Config.Exercises {
+		if exercise.UUID == "" {
+			continue
+		}
+
+		if seen[exercise.UUID] {
+			uuids = append(uuids, exercise.UUID)
+		}
+
+		seen[exercise.UUID] = true
+	}
+
+	return uuids
+}
+
+func duplicateTrackUUID(t track.Track) []string {
+	if disableHTTPChecks {
+		return []string{}
+	}
+
+	// Build up set of uuids to validate.
+	uuids := []string{}
+	for _, exercise := range t.Config.Exercises {
+		if exercise.UUID == "" {
+			continue
+		}
+		uuids = append(uuids, exercise.UUID)
+	}
+
+	payload := struct {
+		TrackID string   `json:"track_id"`
+		UUIDs   []string `json:"uuids"`
+	}{
+		TrackID: t.ID,
+		UUIDs:   uuids,
+	}
+
+	body, err := json.Marshal(payload)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "-> %s\n", err.Error())
+		os.Exit(1)
+	}
+
+	resp, err := http.Post(UUIDValidationURL, "application/json", bytes.NewBuffer(body))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "-> %s\n", err.Error())
+		os.Exit(1)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusConflict {
+		result := struct{ UUIDs []string }{}
+		if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+			fmt.Fprintf(os.Stderr, "-> %s\n", err.Error())
+			os.Exit(1)
+		}
+
+		return result.UUIDs
+	}
+
+	return []string{}
+}
+
 func init() {
 	RootCmd.AddCommand(lintCmd)
+	lintCmd.Flags().BoolVar(&disableHTTPChecks, "no-http", false, "Disable remote HTTP-based linting.")
 }
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -67,6 +67,10 @@ func lintTrack(path string) bool {
 			msg:   "-> The implementation for '%v' is missing a test suite.\n",
 		},
 		{
+			check: missingUUID,
+			msg:   "-> The exercise '%v' was found in config.json, but does not have a UUID.\n",
+		},
+		{
 			check: foregoneViolations,
 			msg:   "-> An implementation for '%v' was found, but config.json specifies that it should be foregone (not implemented).\n",
 		},
@@ -173,6 +177,17 @@ func missingTestSuite(t track.Track) []string {
 			slugs = append(slugs, slug)
 		}
 	}
+	return slugs
+}
+
+func missingUUID(t track.Track) []string {
+	slugs := []string{}
+	for _, exercise := range t.Config.Exercises {
+		if exercise.UUID == "" {
+			slugs = append(slugs, exercise.Slug)
+		}
+	}
+
 	return slugs
 }
 
--- a/cmd/lint_test.go
+++ b/cmd/lint_test.go
@@ -1,6 +1,9 @@
 package cmd
 
 import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"sort"
 	"testing"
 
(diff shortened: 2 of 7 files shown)


cmd/lint.go
```diff
@@ -75,6 +75,10 @@ func lintTrack(path string) bool {
 			check: duplicateSlugs,
 			msg:   "-> The exercise '%v' was found in multiple (conflicting) categories in config.json.\n",
 		},
+		{
+			check: uuidViolations,
+			msg:   "-> The exercise '%v' was found in config.json, but with a (conflicting) UUID.\n",
```
kytrinyx: Will we get two of this message, one for each time it has a conflicting UUID?

I wonder if we should have this message once, showing the UUID not the exercise ID.

cmd/lint.go
```diff
@@ -214,6 +218,22 @@ func duplicateSlugs(t track.Track) []string {
 	return slugs
 }
 
+func uuidViolations(t track.Track) []string {
+	// ps is a set of Slugs partitioned by their UUIDs.
+	ps := map[string][]string{}
```
kytrinyx: How about `uuidsBySlug` as the name for this one?

With maps in Go if I don't have an obvious name, and if it's not just a tiny scope where `m` will be fine, I tend to try to use the `xByY` or `yToX` format (picking whichever seems to work for my context).

kytrinyx (commented):
I'm so glad that you're picking this one up. I don't think I could get to it in the next couple of weeks, and we do really need it!

I have a thought about what to put in the error message, and a small suggestion with respect to a name, but overall I really like where you're taking this.

cmd/lint.go
robphoenix: Could it show the UUID with all duplicated exercises? That would save searching the config.json for the exercises with the duplicate UUID. If that does get unnecessarily complicated then I think returning the duplicated UUID would be preferable also.

robphoenix (commented):
This is looking really good @nywilken, I really like the simple approach you've taken.

cmd/lint.go
kytrinyx: If we do that, we'd need to change the interface of the check() functions, I think.

nywilken: > Will we get two of this message, one for each time it has a conflicting UUID?

Yeah, but I thought about having the uuidViolations function return a concatenated string of slugs and displaying something along the line of 
` The exercises 'a, b' were found in the config.json, but with conflicting UUIDs`


> I wonder if we should have this message once, showing the UUID not the exercise ID.

> Could it show the UUID with all duplicated exercises? 

UUID vs Slug in message works, but to support both I will need to refactor things a bit . TBH, I might not have a choice because to handle UUID violations across multiple tracks we will need to display both UUID and slugs, I think.

I will work on an implementation and submit for feedback. @kytrinyx do you have a URL for the UUID verification API?

cmd/lint.go
nywilken: Thanks for the suggestion and tip. I struggled with the name and eventually threw my hands up in hopes that you or Rob would suggest something better 🙇

nywilken:
> I'm so glad that you're picking this one up. I don't think I could get to it in the next couple of weeks, and we do really need it!

First off let me just say that the amount of features, issues, and reviews you've been wrangling these past few weeks is super impressive.  Thanks for everything you do.

With that being said, I did see some of the open issues pertaining to UUIDs and thought to best get working on it. I hope to have this hashed out sooner rather than later. Configlet for the win 🥇

cmd/lint.go
kytrinyx: I dumped the information about the UUID service here: https://github.com/exercism/configlet/issues/30

Eventually we'll swap that service out with one on the new prototype, but until then we can use that.

kytrinyx: I think it would be fine to have the message be:

```
The UUID %s occurs multiple times. Each exercise must have a unique UUID.
```

That would let people search for the offending UUID.

robphoenix: How about for readability, instead of:

```
The UUID 8d07e3b2-7a94-4cc9-bac1-8bb851b34a8b occurs multiple times. Each exercise must have a unique UUID.
```
We have:
```
This UUID occurs multiple times:  8d07e3b2-7a94-4cc9-bac1-8bb851b34a8b  
Each exercise UUID must be unique.
```
I know it becomes a multiline message, but I think it helps to highlight the cause of the problem from the message, and should make it easier to copy and paste?

robphoenix:
>  the amount of features, issues, and reviews you've been wrangling these past few weeks is super impressive. Thanks for everything you do.

Couldn't agree more @nywilken @kytrinyx :raised_hands:

cmd/lint.go
kytrinyx: Multiline is fine, especially if it helps copy and paste.

How about:

```
The following UUID occurs multiple times. Each exercise UUID must be unique.
8d07e3b2-7a94-4cc9-bac1-8bb851b34a8b
```

That way a triple-click on the uuid will select it.

kytrinyx:
> the amount of features, issues, and reviews you've been wrangling these past few weeks

This would have been _incredibly_ lonely without the two of you for feedback. And the quality would have suffered dramatically. So relieved to have you along for the ride.

nywilken:
The updated lint command validates UUIDs against other tracks and supports the `--no-http` flag, but the connection to the UUID service is, pretty much, still a work in progress. I would like to clean things up a bit and double check the logic/flow.

nywilken:
@kytrinyx @robphoenix this is ready for review - I might change a variable name `results := configErros.check` and `uuids map. ... in missingUUID`. But the functionality is good to go. 

I built the logic for posting to the UUID service directly into the lint command to keep the code lean and to the point. My line of thinking is, if we decide to add more HTTP checks we can build a handler for managing http.Client connections at that point 

Look forward to your review.

robphoenix (approved):

kytrinyx:
> My line of thinking is, if we decide to add more HTTP checks we can build a handler for managing http.Client connections at that point

I think that is a good plan.

cmd/lint.go
```diff
@@ -176,6 +198,26 @@ func missingTestSuite(t track.Track) []string {
 	return slugs
 }
 
+func missingUUID(t track.Track) []string {
+	uuids := map[string]bool{}
```
kytrinyx: The name of this is slightly confusing at first glance since the map contains slugs not uuids. That said, I'm not sure this is worth blocking the PR for. Also I can't think of a great name. The most accurate name I could think of was `slugsToPresenceOfUUID` or `presenceOfUUIDBySlug` but those are pretty wordy.

Let's just think on it for later.

cmd/lint.go
```diff
@@ -213,6 +255,73 @@ func duplicateSlugs(t track.Track) []string {
...
+	for _, exercise := range t.Config.Exercises {
+		if exercise.UUID == "" {
+			continue
+		}
+		uuids = append(uuids, exercise.UUID)
+	}
+
+	payload := struct {
+		Language string   `json:"track_id"`
+		UUIDS    []string `json:"uuids"`
```
kytrinyx: I would name this `UUIDs` not `UUIDS`, since the `s` is not a part of the acronym.

cmd/lint.go
```diff
@@ -213,6 +255,73 @@ func duplicateSlugs(t track.Track) []string {
...
+	var uuids []string
+	for _, exercise := range t.Config.Exercises {
+		if exercise.UUID == "" {
+			continue
+		}
+		uuids = append(uuids, exercise.UUID)
+	}
+
+	payload := struct {
+		Language string   `json:"track_id"`
```
kytrinyx: This should be `TrackID`, not `Language`. If we call this concept language elsewhere, we need to fix that.

cmd/lint.go
```diff
@@ -213,6 +255,73 @@ func duplicateSlugs(t track.Track) []string {
...
+
+		return result.UUIDS
+	}
+
+	return []string{}
+}
+
 func init() {
 	RootCmd.AddCommand(lintCmd)
+	lintCmd.Flags().BoolVar(&disableHTTPChecks, "no-http", false, "skip HTTP based verification checks")
```
kytrinyx: I would typically hyphenate _HTTP-based_, but I don't think that grammar rules are very clear on this one.

cmd/lint_test.go
```diff
@@ -150,3 +153,69 @@ func TestDuplicateSlugs(t *testing.T) {
...
 }
+
+func TestDuplicateUUID(t *testing.T) {
+	tests := []struct {
+		desc     string
+		expected int
+		config   track.Config
+	}{
+		{
+			desc:     "should not complain about a conflicting UUID for exercises with missing UUIDs.",
```
kytrinyx, line 164: Great catch!

cmd/lint_test.go
```diff
@@ -150,3 +153,69 @@ func TestDuplicateSlugs(t *testing.T) {
...
+		},
+		{
+			desc:     "should complain that multiple exercises have a conflicting UUID.",
+			expected: 1,
+			config: track.Config{
+				Exercises: []track.ExerciseMetadata{
+					{Slug: "cherry", UUID: "8d07e3b2-7a94-4cc9-bac1-8bb851b34a8b"},
+					{Slug: "diakon", UUID: "cdb228c9-4682-498f-a87f-f2cf1dd34978"},
+					{Slug: "eggplant", UUID: "8d07e3b2-7a94-4cc9-bac1-8bb851b34a8b"},
+				},
```
kytrinyx: For test data I would typically simplify the UUIDs down to something minimal that makes it easy to scan and understand the problem. So:

```
UUID: "xxx",
UUID: "abc",
UUID: "xxx",
```

cmd/lint_test.go
```diff
@@ -150,3 +153,69 @@ func TestDuplicateSlugs(t *testing.T) {
...
+		uuids := duplicateUUID(track)
+
+		assert.Equal(t, test.expected, len(uuids), test.desc)
+	}
+}
+
+func TestDuplicateTrackUUID(t *testing.T) {
+	fakeEndpoint := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusConflict)
+		fmt.Fprintln(w, `{"uuids": ["c66cfdee-0a93-1c80-a510-4efaa22fca0a6204c6a"]}`)
```
kytrinyx: Ditto here for keeping the fake UUIDs ridiculously simple (and not actually UUIDs).

cmd/lint.go
nywilken: 👍

cmd/lint_test.go
nywilken: Ah that makes complete sense. I'll update the tests and fixtures.

nywilken:
> The name of this is slightly confusing at first glance since the map contains slugs not uuids. 

I simplified the code in `missingUUID`, as it was doing more than it needed to (don't really need a map, nor do I need to check against forgone slugs). I'm much happier with this implementation. 

> This should be TrackID, not Language. If we call this concept language elsewhere, we need to fix that.

Thanks for calling this out. I will update in the lint command for now. But I think we might need to consider adding an additional property to the config.json file for `track_id`.  While testing I noticed that I had to down-case the value in-order for the UUID service to validate properly. Otherwise, all UUIDs are considered duplicates.

I also just came across this [decision thread](https://github.com/exercism/docs/blob/97e2de967f7290967dc0b3754efacc00602101b9/language-tracks/launch/decisions.md#language-name-and-track-id) where I see that the language property does not match the actual track_id. Which means the lint command will report errors for tracks like Cpp and Csharp. 

By adding the `track_id` property maintainers can continue to use `Language` and we could avoid making any transformations to the string on encoding and decoding. This is an easy change, but would like to get the team's thoughts before making changes.

kytrinyx:
> By adding the track_id property maintainers can continue to use Language and we could avoid making any transformations to the string on encoding and decoding. This is an easy change, but would like to get the team's thoughts before making changes.

Yes, let's do that! I can update the UUID service.

> Which means the lint command will report errors for tracks like Cpp and Csharp.

I think I'm not following. What's the error message?

> I will update in the lint command for now. But I think we might need to consider adding an additional property to the config.json file for track_id.

Maybe! If I can understand what we're wrestling with here I might know of some other neat trick we could use.

kytrinyx:
This looks great. Thank you 🌻

kytrinyx (approved):

cmd/lint.go
```diff
@@ -213,6 +246,73 @@ func duplicateSlugs(t track.Track) []string {
...
+			continue
+		}
+		uuids = append(uuids, exercise.UUID)
+	}
+
+	payload := struct {
+		TrackID string   `json:"track_id"`
+		UUIDs   []string `json:"uuids"`
+	}{
+		TrackID: strings.ToLower(t.Config.Language),
```
kytrinyx: It just clicked. This is what you mean. Yes, this is a problem.

Let's assume that the track lives in a directory that is named after the track_id for now.

kytrinyx: I don't know if that will work on CI, though. Looking into that.

kytrinyx: Yes, this will work on Travis.

nywilken: Perfect, this is even easier. Thanks for the quick suggestion and for validating that it will work on Travis.  I'll updated `Track` with an ID field, using the directory name as the track ID. I tested against all tracks - problem resolved.

nywilken:
There was mention of a --uuid-service flag and a possible uuid suggestion for any duplicate uuids found. I didn't get around to working that into this PR, but I can revisit later if needed. I'll merge once the build goes green.

kytrinyx:
I think that for now we don't need those. Let's just keep them in mind if not having them becomes painful.

PR #59, opened 2017-08-21
Replace calls to Fprintf with Fprintln

nywilken:
The use of fmt.Fprintf should be reserved for outputting messages that require string interpolation or formatting that can not be achieved with fmt.Fprintln.

This change also includes a minor tweak to the version command's output.


robphoenix (approved):

kytrinyx (approved):
Good call.

nywilken:
Thanks for the review. I'll merge once the build goes green.

PR #63, opened 2017-08-23
fmt: fix topic normalisation

robphoenix:
See https://github.com/exercism/meta/issues/83

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -145,7 +146,17 @@ func formatTopics(m map[string]interface{}) {
 
 func normaliseTopic(t string) string {
 	s := strings.ToLower(t)
-	s = strings.Replace(s, " ", "-", -1)
+
+	// we only want to let through letters and underscores
+	// for the final output.
+	// hyphens and whitespace are allowed through for now
+	// as word delimiters to be replaced with underscores.
+	reg := regexp.MustCompile("[^a-z\\s-_]+")
+	s = reg.ReplaceAllString(s, "")
+
+	// output to be snake_case
+	reg = regexp.MustCompile("[\\s-]+")
+	s = reg.ReplaceAllString(s, "_")
 	return s
 }
 
--- a/fixtures/format/malformed/config.json
+++ b/fixtures/format/malformed/config.json
@@ -4,7 +4,7 @@
       "core": false,
       "difficulty": 1,
       "slug": "one",
-      "topics": [ "integers", "Control-Flow (conditionals)", "Booleans", "logic" ],
+      "topics": [ "integers", "Control-Flow    (conditionals)", "Booleans", "logic" ],
       "unlocked_by": null,
       "uuid": "001"
     },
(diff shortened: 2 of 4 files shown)


cmd/fmt.go
```diff
@@ -145,7 +146,19 @@ func formatTopics(m map[string]interface{}) {
...
 func normaliseTopic(t string) string {
 	s := strings.ToLower(t)
-	s = strings.Replace(s, " ", "-", -1)
+
+	// we only want letters, hyphens and
+	// whitespace (to be replaced) in topics
+	reg := regexp.MustCompile("[^a-zA-Z\\s-]+")
+	s = reg.ReplaceAllString(s, "")
+
+	// corner case where prefix should be removed
```
robphoenix: This hasn't been confirmed yet, but I thought I'd put it in here for now, just wanted to consider how it would be done.

kytrinyx: Sounds good. I'll try to push all of this to be confirmed today.

cmd/fmt.go
```diff
@@ -145,7 +146,17 @@ func formatTopics(m map[string]interface{}) {
 
 func normaliseTopic(t string) string {
 	s := strings.ToLower(t)
-	s = strings.Replace(s, " ", "-", -1)
+
+	// we only want to let through letters and underscores
+	// for the final output.
+	// hyphens and whitespace are allowed through for now
+	// as word delimiters to be replaced with underscores.
+	reg := regexp.MustCompile("[^a-zA-Z\\s-_]+")
```
nywilken: There's a call to ToLower above so the regex could be simplified to `[^a-z\\s-_]+`. Or even `(?i)[^a-z\\s-_]+` if you prefer the regex to be case insensitive.

cmd/fmt.go
```diff
@@ -145,7 +146,17 @@ func formatTopics(m map[string]interface{}) {
...
+	s = reg.ReplaceAllString(s, "")
+
+	// output to be snake_case
+	s = strings.Replace(s, " ", "_", -1)
```
nywilken: If there is a topic that has a double space or double dash (user typo) the resulting string will have a double underscore. What do you think about using another regex replace here with the pattern `[\\s-]+`?

nywilken (commented):
@robphoenix this looks good - just a few suggestions  on the string replacements.

cmd/fmt.go
robphoenix: ah yes, of course, thanks!

cmd/fmt.go
robphoenix: This is a much nicer way of doing it, avoiding multiple calls and covering more cases. :pray:

fixtures/format/malformed/config.json
```diff
@@ -15,7 +15,7 @@
       "topics": [
 "equality",
              "mathematics",
-                    "text-formatting",
+                    "text--formatting",
```
robphoenix, line 18: I updated the malformed test case to reflect the improvements suggested by @nywilken

nywilken (approved):

nywilken:
@kytrinyx @robphoenix this looks good and I would like to merge it if there are no any pending items being discussed?

robphoenix:
It's good to go from my point of view.

kytrinyx (approved):
Let's ship this!

nywilken:
@robphoenix I missed your response. Thanks for merging.

PR #64, opened 2017-08-24
Ammendments to commands

robphoenix:
Somehow I missed `cobra.ExactArgs` and so have changed
`cobra.MinimumNArgs` to this where it was warranted

Along the way I noticed the error output formatting was inconsistent,
so I brought it inline with the `lint` command as that was implemented
first. I also added a path exists check to `lint`.

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -29,7 +29,7 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
 `,
 	Example: fmt.Sprintf("  %s fmt %s --verbose", binaryName, pathExample),
 	Run:     runFmt,
-	Args:    cobra.MinimumNArgs(1),
+	Args:    cobra.ExactArgs(1),
 }
 
 // formatter applies additional formatting to unmarshalled JSON files.
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -55,12 +55,12 @@ func runFmt(cmd *cobra.Command, args []string) {
 
 	for _, f := range fs {
 		if _, err := os.Stat(f.path); os.IsNotExist(err) {
-			fmt.Fprintln(os.Stderr, err.Error())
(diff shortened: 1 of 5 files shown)


cmd/fmt.go
```diff
@@ -55,30 +55,30 @@ func runFmt(cmd *cobra.Command, args []string) {
...
 	for _, f := range fs {
 		if _, err := os.Stat(f.path); os.IsNotExist(err) {
-			fmt.Fprintln(os.Stderr, err.Error())
+			fmt.Fprintf(os.Stderr, "-> path not found: %s\n", f.path)
 			os.Exit(1)
 		}
 		diff, formatted, err := formatFile(f.path, f.formatter)
 		if err != nil {
-			fmt.Fprintln(os.Stderr, err.Error())
+			fmt.Fprintf(os.Stderr, "-> %s", err.Error())
```
nywilken, line 63: It's a good idea to keep the message formatting the same for messaging generated by configlet commands (line 58), but I'm on the fence about errors returned from other packages, as we don't actually control the messaging.  

With that being said, maybe having a printer.go in  configlet that we could use to print lines with the correct format would be worth looking into. It will reduce code changes in the future, help with testing, and keep me from forgetting to add the "->" 😉

kytrinyx: I think consistency is worth it even if we don't control the messaging.

kytrinyx (approved):
This looks great. Agreed on maybe adding a helper eventually to help with outputting the final errors nicely to the user.

nywilken (approved):

cmd/fmt.go
robphoenix: > With that being said, maybe having a printer.go in configlet that we could use to print lines with the correct format would be worth looking into.

Yes, we should separate any UI particulars from the actual code. We're just gonna keep forgetting those `->` otherwise :smile: 

If you know of any good examples of this, I'd be interested to see, and look into implementing this in due course.

PR #67, opened 2017-09-04
Update track id detection to use absolute path

nywilken:
In adding a track.ID property an issue was introduced where configlet does not properly set a track's ID when running `configlet lint .` within a track's root directory. 

This change ensures that filepath.Base is called on a absolute path, as opposed to a relative path.

### DIFF
--- a/track/track.go
+++ b/track/track.go
@@ -19,7 +19,12 @@ func New(path string) (Track, error) {
 	track := Track{
 		path: filepath.FromSlash(path),
 	}
-	track.ID = filepath.Base(path)
+
+	ap, err := filepath.Abs(track.path)
+	if err != nil {
+		return track, err
+	}
+	track.ID = filepath.Base(ap)
 
 	c, err := NewConfig(filepath.Join(path, "config.json"))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


nywilken:
@kytrinyx @robphoenix I introduced a bug with my latest UUID duplicate check commit. I would like to get your eyes on my fix before merging. Thanks!

kytrinyx (approved):
This looks right to me.

robphoenix (approved):

nywilken:
Thanks for the review - my bad on breaking configlet.

kytrinyx:
Do we need a new release?

nywilken:
Yes, but I think we need to wait until https://github.com/exercism/bash/pull/83 is merged.

kytrinyx:
Ok, the bash track is sorted out.

Do you have access to make configlet releases? (And have I documented it well enough that you could?)

nywilken:
Sweet! I'm not sure about the release. It seems like I have perms to edit previous releases, and draft a new one. So I just might. 

If you shoot me what docs you have I can try releasing later today - I'm currently away from my  machine.

kytrinyx:
There was no documentation (of course /facepalm) so I added some: https://github.com/exercism/configlet/pull/68

If you get stuck trying to follow it, holler, and we can figure out what is missing.

PR #72, opened 2017-09-10
Implement UI printer

### LINKED ISSUE #66: Implement a UI printer
Currently error messages are printed to stdout/stderr with a `-> ` prefix, which is individually included in each `print` call. As suggested by @nywilken [here](https://github.com/exercism/configlet/pull/64#discussion_r135339064) we should offload this to a separate UI printer mechanism.

robphoenix:
Closes #66

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/exercism/configlet/ui"
 	"github.com/pmezard/go-difflib/difflib"
 	"github.com/spf13/cobra"
 )
@@ -56,12 +57,12 @@ func runFmt(cmd *cobra.Command, args []string) {
 
 	for _, f := range fs {
 		if _, err := os.Stat(f.path); os.IsNotExist(err) {
-			fmt.Fprintf(os.Stderr, "-> path not found: %s\n", f.path)
+			ui.PrintError("path not found:", f.path)
 			os.Exit(1)
 		}
 		diff, formatted, err := formatFile(f.path, f.formatter)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "-> %s", err.Error())
+			ui.PrintError(err.Error())
 			continue
 		}
 		if diff == "" {
@@ -69,17 +70,17 @@ func runFmt(cmd *cobra.Command, args []string) {
 		}
 		err = ioutil.WriteFile(f.path, formatted, os.FileMode(0644))
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "-> %s", err.Error())
+			ui.PrintError(err.Error())
 			continue
 		}
 		if fmtVerbose {
-			fmt.Printf("-> %s\n\n%s\n", f.path, diff)
+			ui.Print(f.path, "\n\n", diff)
 		}
 		changes += fmt.Sprintf("%s\n", f.path)
 	}
 
 	if changes != "" {
-		fmt.Printf("-> changes made to:\n%s", changes)
+		ui.Print("changes made to:\n", changes)
 	}
 	return
 }
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -8,6 +8,7 @@ import (
 	"os"
 
 	"github.com/exercism/configlet/track"
+	"github.com/exercism/configlet/ui"
 	"github.com/spf13/cobra"
 )
 
@@ -47,12 +48,12 @@ func runLint(cmd *cobra.Command, args []string) {
 
 func lintTrack(path string) bool {
 	if _, err := os.Stat(path); os.IsNotExist(err) {
-		fmt.Fprintf(os.Stderr, "-> path not found: %s\n", path)
+		ui.PrintError("path not found:", path)
 		os.Exit(1)
 	}
 	t, err := track.New(path)
 	if err != nil {
-		fmt.Printf("-> %s\n", err)
+		ui.PrintError(err.Error())
 	}
 
 	configErrors := []struct {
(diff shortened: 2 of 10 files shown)


robphoenix:
I have to fix the failing build tomorrow.

cmd/fmt.go
```diff
@@ -56,30 +57,30 @@ func runFmt(cmd *cobra.Command, args []string) {
 
 	for _, f := range fs {
 		if _, err := os.Stat(f.path); os.IsNotExist(err) {
-			fmt.Fprintf(os.Stderr, "-> path not found: %s\n", f.path)
+			ui.Error("path not found: " + f.path)
```
nywilken: What do you think about calling this `ui.PrintError`?  

`ui.Error` doesn't seem to provide enough context on what it does.

ui/printer.go
```diff
@@ -0,0 +1,30 @@
...
+var (
+	// prefix is prepended to ui messages
+	prefix = "->"
+	// ErrOut is a writer for error messages
+	ErrOut io.Writer = os.Stderr
+	// PrintOut is a writer for normal messages
+	PrintOut io.Writer = os.Stdout
+)
+
+// Print prints msg to standard output
```
nywilken: Maybe `// Print writes msg to ui.PrintOut` since PrintOut can be any io.Writer

ui/printer.go
```diff
@@ -0,0 +1,30 @@
...
+func Print(msg string) {
+	stdPrinter(PrintOut, msg)
+}
+
+// Error prints msg to standard error
```
nywilken: Same for ErrorOut

cmd/fmt.go
```diff
@@ -56,30 +57,30 @@ func runFmt(cmd *cobra.Command, args []string) {
...
-			fmt.Printf("-> %s\n\n%s\n", f.path, diff)
+			ui.Print(f.path + "\n\n" + diff)
 		}
 		changes += fmt.Sprintf("%s\n", f.path)
 	}
 
 	if changes != "" {
-		fmt.Printf("-> changes made to:\n%s", changes)
+		ui.Print("changes made to:\n" + changes)
 	}
```
nywilken: This look good. Have you consider making `ui.Print` a variadic function? So that we can print without having to concat the string before hand.

ui/printer_test.go
```diff
@@ -0,0 +1,22 @@
+package ui
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPrint(t *testing.T) {
```
nywilken: 👍

nywilken (commented):
I took a first pass at this PR; I like the approach. I left a couple of questions and suggestions. Let me know what you think.

cmd/fmt.go
robphoenix: :+1:

cmd/fmt.go
robphoenix: I did, I tried it and get stumped by types. But, at this suggestion I tried again, and got it to work. So thankyou for making me go back and figure it out! :relieved:

ui/printer.go
robphoenix: :+1:

ui/printer.go
robphoenix: :+1:

ui/printer_example_test.go
```diff
@@ -0,0 +1,17 @@
...
+import "os"
+
+func ExamplePrint() {
+	Out = os.Stdout
+	Print("this is a message")
+	// Output:
+	// -> this is a message
+}
+
+func ExampleError() {
```
nywilken: This should be renamed to `ExamplePrintError()` for consistency purposes.

ui/printer_test.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func TestPrintMultipleStrings(t *testing.T) {
+	var buf bytes.Buffer
+	Out = &buf
+	Print("this", "is", "a", "message")
+	assert.Equal(t, "-> this is a message\n", buf.String())
+}
+
+func TestError(t *testing.T) {
```
nywilken: This should also be renamed for consistency `TestPrintError`

ui/printer_test.go
```diff
@@ -0,0 +1,36 @@
...
+	var buf bytes.Buffer
+	ErrOut = &buf
+	PrintError("this is an error")
+	assert.Equal(t, "-> this is an error\n", buf.String())
+}
+
+func TestErrorMultipleStrings(t *testing.T) {
```
nywilken: Just thinking out loud, but you could consolidate these into two table tests (i.e TestPrint and TestPrintError).

ui/printer_test.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPrint(t *testing.T) {
+	var buf bytes.Buffer
+	Out = &buf
```
nywilken: I know this is a simple test, but we should restore the original value of Out once the test is done. This is a good habit to get into to avoid issues with tests later on. 

```
saved := Out
Out = &buf
defer func() { Out = saved }()
```

nywilken (commented):
@robphoenix glad you were able to get the variadic function working. This looks good - just a few nitpicks.

ui/printer_example_test.go
robphoenix: good catch!

ui/printer_test.go
robphoenix: good advice, thanks.

ui/printer_test.go
robphoenix: done :heavy_check_mark:

robphoenix:
Thanks for the review @nywilken. I've updated on your suggestions, the code is better for them :100:.

nywilken (approved):
This looks good to me. Feel free to merge when you're ready.

cmd/lint.go
```diff
@@ -47,12 +48,12 @@ func runLint(cmd *cobra.Command, args []string) {
...
 func lintTrack(path string) bool {
 	if _, err := os.Stat(path); os.IsNotExist(err) {
-		fmt.Fprintf(os.Stderr, "-> path not found: %s\n", path)
+		ui.PrintError("path not found:", path)
 		os.Exit(1)
 	}
 	t, err := track.New(path)
 	if err != nil {
-		fmt.Printf("-> %s\n", err)
+		ui.PrintError(err.Error())
```
nywilken: @robphoenix this should exit with a status 1, or return true, so that the lint command terminates with an exit status of 1.

nywilken: Sorry didn't mean to impose here. If you planned on merging this soon, would it be okay to include this fix in your PR?

nywilken: Sorry I have a test in mind that I would like to add for this issue so let's keep things within scope and not make the change here. I'll have time tonight to address it. Cheers!

PR #78, opened 2017-09-17
lint: Fix error handling for malformed config files

### LINKED ISSUE #77: configlet lint: should it exit non-zero if config.json is unparseable as JSON? (it currently exits zero)
https://github.com/exercism/ceylon/pull/50 introduces obviously invalid JSON in config.json (missing comma between properties of an object, and an unclosed opening brace)

In the build, https://travis-ci.org/exercism/ceylon/builds/276065890 we can see

```
$ bin/configlet lint .
-> invalid config config.json -- invalid character '{' after object key:value pair
The command "bin/configlet lint ." exited with 0.
```

configlet has correctly determined that the file is invalid JSON. But then it exits zero. Should it be nonzero instead, so we can be assured that it will only exit zero if the config is valid JSON that also passes lints? It would be helpful for CI if we can be assured this is the case.

nywilken:
This will ensure that a zero exit status code is only returned for a valid track that has passed all lint checks.

closes #77

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -49,11 +49,13 @@ func runLint(cmd *cobra.Command, args []string) {
 func lintTrack(path string) bool {
 	if _, err := os.Stat(path); os.IsNotExist(err) {
 		ui.PrintError("path not found:", path)
-		os.Exit(1)
+		return true
 	}
+
 	t, err := track.New(path)
 	if err != nil {
 		ui.PrintError(err.Error())
+		return true
 	}
 
 	configErrors := []struct {
@@ -98,7 +100,7 @@ func lintTrack(path string) bool {
 		},
 	}
 
-	hasErrors := false
+	var hasErrors bool
 	for _, configError := range configErrors {
 		failedItems := configError.check(t)
 
--- a/cmd/lint_test.go
+++ b/cmd/lint_test.go
@@ -4,12 +4,42 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"path/filepath"
 	"sort"
 	"testing"
 
 	"github.com/exercism/configlet/track"
 	"github.com/stretchr/testify/assert"
 )
+
+func TestLintTrack(t *testing.T) {
+	lintTests := []struct {
+		desc     string
+		path     string
+		expected bool
+	}{
+		{
+			desc:     "should fail when given a track containing one or more lint failures.",
+			path:     "../fixtures/numbers",
+			expected: true,
+		},
+		{
+			desc:     "should fail when given a track containing malformed configuration data.",
+			path:     "../fixtures/broken-maintainers",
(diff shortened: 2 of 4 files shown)


robphoenix (approved):
:+1:

nywilken:
@robphoenix thanks again for the review.

stkent:
Is something similar required for `fmt`?

https://github.com/exercism/configlet/blob/c4b48902011c10db66e9803db7fcf52bd98d324d/cmd/fmt.go#L58-L85

Looks like we get a non-zero exit status if the path is incorrect, but not if e.g. there's a formatting error?

stkent:
I stumbled across this because `fmt` fixed some topic names for me, but `lint` wasn't reporting any issues. Maybe the solution is instead to add `lint` checks for formatting issues? That would avoid needing to manually check new topic adds in PRs, for example.

nywilken:
> Looks like we get a non-zero exit status if the path is incorrect, but not if e.g. there's a formatting error?

What do you mean by formatting error? An error that can not be fixed via the fmt command? Or a formatting error that gets repaired via `fmt`? If its the latter, then the `fmt` wouldn't error out, as it is expected to work like [gofmt](https://golang.org/cmd/gofmt/) and fix file formatting. 

Ideally, maintainers would run `fmt` after making changes to the config.json or maintainers.json files to fix any formatting issues. But if topics are causing issues then maybe we can think about adding a flag to `fmt` that can be part of the track build. Or as you suggested updating the lint command. 

Would you mind opening up a new issue with some insight on how you would expect the lint command to behave when it comes to reporting formatting issues?

stkent:
> What do you mean by formatting error? An error that can not be fixed via the fmt command? Or a formatting error that gets repaired via `fmt`? If its the latter, then the `fmt` wouldn't error out, as it is expected to work like gofmt and fix file formatting.

Ah, this makes sense, thanks for explaining! I did indeed mean "a formatting error that gets repaired via `fmt`".

> Ideally, maintainers would run `fmt` after making changes to the config.json or maintainers.json files to fix any formatting issues.

The issue I'm running into occurs when maintainers accept PRs from non-maintainers. This (fortunately) happens a lot on the Java track, but also means there are plenty of opportunities for the track config to 'drift' (because `lint` doesn't fail our CI builds if the config file formatting isn't up to spec). Right now I manually re-run `fmt` locally every so often, and this may well be a better long-term solution than making the CI build a lot more intolerant => hostile towards external contributors, especially since the `configlet` install process is non-trivial.

I'm happy to open a new discussion issue if you'd still like to explore this further in light of my replies here; let me know!

kytrinyx:
I want to write a bot that submits PRs to PRs to fix formatting issues. We shouldn't have to deal with this manually. I have a lot on my plate at the moment, though, so I won't get it it very soon.

Issue #33, opened 2017-07-27
Possible graph/tree command


robphoenix:
@jonmcalder has done some [awesome work](https://github.com/exercism/discussions/issues/175#issuecomment-318243990) visualizing the graph of exercise dependencies in a `config.json`.

I wonder if it would be beneficial to add a graph command that could output something similar.

As Jon has used R it would have to be recreated in Go. I imagine this could be done by looping over the slice of `ExerciseMetadata` in a `Config` struct and building up the graph. This could then be used to possibly either output  a [graphviz](http://graphviz.org/About.php) dot file, using [this library](https://github.com/awalterschulze/gographviz) or building a web page, perhaps using [d3.js](https://d3js.org/)?

kytrinyx:
I think this would be tremendously valuable.

robphoenix:
Reading @iHiD's [comment](https://github.com/exercism/discussions/issues/175#issuecomment-318368925) perhaps a simple terminal output similar to Linux's `tree` command might be better?

```
.
├── hello-world
│   ├── leap
│   ├── gigasecond
│   └── clock
├── bob
├── beer-song
├── luhn
└── meetup
    ├── hamming
    ├── grains
    │   ├── raindrop
    │   └── forth
    └── house

13 exercises, 5 core
```

Would probably be easier to implement than what I've outlined above.

kytrinyx:
A tree-like view would probably be easier to implement—and I think it would probably be a good, understandable visual, as well.

robphoenix:
Okay, grand, I'll get on it.

tleen:
:laughing:  Now that we are actually starting to analyze the layout of the Go track I came here to suggest this as an option for configlet. A co-maintainer beat be to it!

Looks like configlet may be becoming a more general purpose config-analyzer. If so I believe another helpful view of config would be a listing (and count) of exercises by difficulty.

kytrinyx:
> If so I believe another helpful view of config would be a listing (and count) of exercises by difficulty.

Yeah, Configlet is turning into our all-purpose track maintenance tool. Go ahead and open up the issue for the exercise list, and we can dig into the details of it.

tleen:
Will do. It may be a nice option in the tree view to have the difficulty next to the exercise:

```text
└── meetup (3)
    ├── hamming (5)
    ├── grains (7)
```

robphoenix:
This is a great idea @tleen!

PR #83, opened 2017-09-30
Visualization

tleen:
This will add a visualization command for a tree based output of a configuration file (#33). Initially prototyped in [exercism-config-tree](https://github.com/tleen/exercism-config-tree) but properly belonging to configlet.

This is my first shot at a Cobra based command line app so feedback would be greatly appreciated!


cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+var treeSpacing = strings.Repeat(" ", indent)
+
+var treeBranching = strings.Repeat(branch, indent-1)
+
+// showDifficulty holds --difficulty flag value indicates that we
+// should display exercise difficulty after slug, by default we do not.
+var showDifficulty bool
+
+// visCmd defines the visualize command
+var visCmd = &cobra.Command{
```
kytrinyx: I would typically name the prefix the same as the file, so `visualizeCmd`

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+	Example: fmt.Sprintf("  %s visualize %s --difficulty", binaryName, configPathExample),
+	Run:     runVisualizer,
+	Args:    cobra.ExactArgs(1),
+}
+
+// slugToExercise is a global lookup table of slugs to exercises.
+var slugToExercise = map[string]*exerciseUnlocks{}
+
+// exerciseUnlocks is an extension of the exercise metatata type with with
+// the exercises it unlocks.
```
kytrinyx: _Unlocks_ is a bit ambiguous, since it looks like it's a verb, but it's being used as a noun. How about something like "dependentExercise" or "childExercise" or something like that?

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+type exerciseUnlocks struct {
+	track.ExerciseMetadata
+	Unlocks []string // slugs of unlocked exercises
```
kytrinyx: I would probably use `SomethingSlugs` here, where `Something` is whatever you end up naming the type... `ChildSlugs` or `DependentSlugs` or some better name that I've not thought of.

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// getDescription is a utility that will return the description for
+// an exerciseUnlock with the difficulty appended if the --difficulty
+// flag was set.
+func (e exerciseUnlocks) getDescription() string {
```
kytrinyx: I'm wary of names that start with `get`. I like to name functions that return a value after _what they are_ not _what they do_. Could this be just `description()`?

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+		buffer.WriteString(fork)
+	}
+
+	// Show the exercise name, it will have the standard branch prefix.
+	buffer.WriteString(treeBranching)
+	buffer.WriteString(" ")
+	buffer.WriteString(e.getDescription())
+	writeln(buffer.String())
+
+	// Now go into the children unlocks and do this all over again.
```
kytrinyx: I'm liking "children" as a name, since you naturally use this when talking about it in comments.

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+	// ...except for the last element because there is nothing below it
+	// to space out.
+	if depth == 0 && !isLast {
+		writeln(trunk)
+	}
+}
+
+func visualizeTrack(path string) error {
+
+	// exercises is a list of all non-deprecated exercises, in config order.
```
kytrinyx: ✂️ the initial empty line, I think, since none of the other functions start with an empty line.

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// Print a header: the language name with markdown style h1 underlining.
+	writeln(config.Language)
+	writeln(strings.Repeat("=", utf8.RuneCountInString(config.Language)))
+
+	// Initial scan through the exercises for this track: filter out deprecated
+	// exercises and setup the data structures from above.
+	for _, e := range config.Exercises {
+		// Completely ignore deprecated exercises. They are dead to us.
```
kytrinyx: 😄 yes, they are dead to us

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+		if e.IsDeprecated {
+			continue
+		}
+		// Container for this exercise
```
kytrinyx: This is getting super nitpick-y, but I like to have complete sentences for comments, so I would say _Create a container for this exercise._ (including punctuation).

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+		eu := exerciseUnlocks{
+			e,
+			make([]string, 0), // Our unlock slugs, filled in on second pass.
+		}
+
+		exercises = append(exercises, eu)
+
+		// Add to slug based global lookup table
```
kytrinyx: (nitpick: punctuation)

cmd/visualize.go
```diff
@@ -0,0 +1,275 @@
...
+		if e.UnlockedBy == "" {
+			continue
+		}
+
+		unlocksPresent = true
+		parent := slugToExercise[e.UnlockedBy]
+		parent.Unlocks = append(parent.Unlocks, e.Slug)
+	}
+
+	// This is more of a warning than an error. No stdwarn :(
```
kytrinyx: It sounds like it should still go to STDERR. My personal guideline for STDOUT vs STDERR is, if I'm piping the output to a file, would I want it to be included by default?

kytrinyx (changes requested):
This looks great! I have a few minor comments, but the overall direction and implementation look super solid. 🌻 🌻 🌻

cmd/visualize.go
tleen: Yeah I was a little uncomfortable with the name. I'll go whole hog into the Parent/Child terminology, it makes sense in this context.

cmd/visualize.go
tleen: Right. I'm sending all the warnings to stderr via `ui.PrintError` to keep the output file clean.

tleen:
Thanks @kytrinyx these were good suggestions!

cmd/visualize.go
kytrinyx: That sounds good.

cmd/visualize.go
```diff
@@ -0,0 +1,273 @@
...
+	track.ExerciseMetadata
+	// childSlugs will be used to store the slugs of exercises unlocked by this parent.
+	childSlugs []string
+}
+
+// description is a utility that will return the description for
+// an exerciseUnlock with the difficulty appended if the --difficulty
+// flag was set.
+func (e exerciseParent) description() string {
+
```
kytrinyx: ✂️ let's ditch the blank line on this one as well.

kytrinyx (approved):
This looks great. I like how this ended up with the new names. I have one last suggested blank line deletion (sorry!), but after that this is good to be merged.

cmd/visualize.go
```diff
@@ -0,0 +1,272 @@
...
+// configurationWarning is boilerplate that will be appended to any warning
+// regarding potential mis-configuration for nextercism.
+const configurationWarning = ", this track may be missing a nextercism compatible configuration."
+
+// configPathExample is used in the Cobra usage output where the configfile path
+// would be.
+const configPathExample = "<path/to/config.json>"
+
+// treeSpacing and treeBranching will be user over and over again
```
robphoenix: s/user/used

cmd/visualize.go
```diff
@@ -0,0 +1,272 @@
...
+// in the tree creation, so create them once here.
+var treeSpacing = strings.Repeat(" ", indent)
+
```
robphoenix: delete this blank line maybe? As the comment refers to both vars? Not sure about this really, no biggie.

cmd/visualize.go
```diff
@@ -0,0 +1,272 @@
...
+var treeBranching = strings.Repeat(branch, indent-1)
+
+// showDifficulty holds --difficulty flag value indicates that we
+// should display exercise difficulty after slug, by default we do not.
```
robphoenix: minor rewording

```
// showDifficulty holds the --difficulty flag value to indicate whether we
// should display the exercise difficulty after slug, by default we do not.
```

cmd/visualize.go
```diff
@@ -0,0 +1,272 @@
...
+├── hello-world
+│
+├── hamming
+│   ├── nucleotide-count
+│   └── rna-transcription
+...
+
+`,
+	Example: fmt.Sprintf("  %s visualize %s --difficulty", binaryName, configPathExample),
+	Run:     runVisualizer,
```
robphoenix: To keep in sync with the rest of the commands, and the use of the word `visualize` int his command, I'd rename this `runVisualize`

cmd/visualize.go
```diff
@@ -0,0 +1,272 @@
...
+	Args:    cobra.ExactArgs(1),
+}
+
+// slugToExercise is a global lookup table of slugs to exercises.
+var slugToExercise = map[string]*exerciseParent{}
+
+// exerciseParent is an extension of the exercise metatata type with with
```
robphoenix: duplicate `with`

cmd/visualize.go
```diff
@@ -0,0 +1,272 @@
...
+
+// writeln is a convenience wrapper around printing the line directly
+// in case we (eventually) need to do something else with the string
+// or writer beforehand.
+func writeln(s string) {
+	fmt.Fprintln(os.Stdout, s)
+}
+
+// writelns will write an array of lines using writeln.
+func writelns(ss []string) {
```
robphoenix: you could change this to a variadic function and get rid of `writeln`

```
func writelns(ss ...string)
```

you then wouldn't need to wrap multiple strings in a slice

```
writelns("", "core", "----")
```

cmd/visualize.go
```diff
@@ -0,0 +1,272 @@
...
+			ui.PrintError(err)
+		}
+	}
+}
+
+// printConfigurationWarning is a utility that will print s to Error,
+// but will prefix it with a nextercism specific configuration warning
+// use it if you expect something via nextercism but it is not present in
+// the configuration.
+func printConfigurationWarning(s string) {
```
robphoenix: should this go in the `ui` package?

cmd/visualize.go
```diff
@@ -0,0 +1,272 @@
...
+	// bonusExercises are slugs of exercises determined to be bonuses:
+	// non-core, no unlocks, in config order.
+	bonusExercises := make([]string, 0)
+
+	config, err := track.NewConfig(path)
+	if err != nil {
+		return err
+	}
+
+	// Print a header: the language name with markdown style h1 underlining.
```
robphoenix: I wonder if we need the header? The user should be aware of which track they are visualizing, no? No biggie.

robphoenix (commented):
This is excellent @tleen, so happy to see this PR from you! :heart: 

I have a few comments and questions.

I wonder if it would be better to give the path to the track as the arg to the command, rather than specifying the `config.json` file. We can add on the filename, and it's less work for the user, not having to specify something that is a given, we know the `config.json` file holds the info, we just need to know where it is. It also means the command is less specific if the track structure changes and we need to reference more files or something.

I love the word `visualize`, but I'm concerned it's too wieldy for a command name. It's quite long and being British I tend to spell it `visualise` (https://ell.stackexchange.com/questions/5741/visualization-or-visualisation). I worry that it's appropriateness may be outweighed by it getting a little annoying to type. Alternatives could be `view`, `tree`, `show`, `list`.

The implementation though is top notch :100:

tleen:
> I wonder if it would be better to give the path to the track as the arg to the command, rather than specifying the config.json file. 

I thought about this some before building it to the filename. I looked at the lint command and did see that it uses a path-to-track. But then I thought that lint's concern was to lint the entire track so that made sense. The concern here is to visualize the config.json file, and really the track could have nothing to do with it. I've downloaded configfiles directly just to visualize them without having the whole track. Even though configlet does concern itself with tracks, I though that level of consistency with other command may take away some of the flexibility in visualizing configurations. Its kind of a toss up. it could really go either way, I don't feel strongly about using the path-to-file.

tleen:
> I love the word visualize, but I'm concerned it's too wieldy for a command name. I

Yes it may be. Just typing out visualize started to annoy the Unix in me. It would be a good change to shorten and get rid of a alternatively spelled word. Leaning towards `tree` since this is a track version of the unix `tree` command. I'll make this change if nobody thinks it would get confusing or has a better name!

cmd/visualize.go
tleen: This is such a better idea!

cmd/visualize.go
tleen: What do you think? I would probably still need this wrapper here because I use configuration warning boilerplate. Really I wanted to wrap up something that was like PrintError but was for PrintWarning, but since it would still be writing to Error it seemed more appropriate to keep it local. But I wanted it to be clear in the code that it was *not* an error. 

Do you think there is a place in ui for a boilerplated warning about configuration like this?

cmd/visualize.go
tleen: The user should know since they need to give the path to it. The header is more of a convenience for the user since I see this being outputted to a file often then pasted into comments in track configuration discussions. I almost want to put some sort of commit hash stamp in there too! 

Without making things too complicated it make make sense (if not here, soon) to add a more markdown compatible mode for copying and pasting into comments. We would have to wrap the actual tree in a `text` code block which would not look great on the command line. Perhaps --md or something? That would add the adornments.

cmd/visualize.go
tleen: I'll fold this one into the command rename.

cmd/visualize.go
tleen: It should probably be called writeLines too.

cmd/visualize.go
robphoenix: I think it's doing a similar thing to `ui.Print` and `ui.PrintError`, wrapping a message to the user in a consistent manner.

> I would probably still need this wrapper here because I use configuration warning boilerplate. 

I'm not sure I follow. Do you need this code in the `cmd` package? Is `configurationWarning` only used by `printConfigurationWarning`? It could be inside the function, rather than a global.

This could go in the `ui` package as a `ui.PrintConfigWarning`. It feels like an appropriate separation of concerns. I think that would still keep it clear in the code that it's not an error. I could be wrong though. :smiley:

cmd/visualize.go
robphoenix: >  I see this being outputted to a file often then pasted into comments in track configuration discussions

Totally valid point, yes you're right. :+1: 

> Without making things too complicated it make make sense (if not here, soon) to add a more markdown compatible mode for copying and pasting into comments. We would have to wrap the actual tree in a text code block which would not look great on the command line. Perhaps --md or something? That would add the adornments.

Yes, I think this is certainly a possibility.

robphoenix:
> The concern here is to visualize the config.json file, and really the track could have nothing to do with it. I've downloaded configfiles directly just to visualize them without having the whole track. 

This is a really good point. If we wanted I suppose we could even allow both by including a check on the arg, if it's a directory append the filename? This would take away the either/or choice and let the command do the work rather than the user.

robphoenix:
> Leaning towards tree since this is a track version of the unix tree command. I'll make this change if nobody thinks it would get confusing or has a better name!

:+1:

cmd/visualize.go
tleen: Yeah I could hardcode the `configurationWarning` var into the `printConfigurationWarning`. It seemed like one of those things that you would put on top of a file as a easy to find setting for the code. That may be me coding Go like another language instead of doing things idomatically. I could move the whole thing over to ui if we think other commands could find it useful. In the tree implementation it is very much centered around nextercism which is why the configuration warning boilerplate is about nextercism. That would probably be appropriate for all warnings.

@nywilken what do you think? Could you see a use for a `ui.PrintConfigurationWarning`? or a `u.PrintWarning`? If we had that we could then have a global flag to enable/disable warnings...

(This may be more change than we should do in the context of this command being added?)

robphoenix: >  I could move the whole thing over to ui if we think other commands could find it useful.

In my mind it's not necessarily about making it accessible to other commands, it's more that it's functionality is inline with the `ui` methods. I fear I may be overthinking this.

> This may be more change than we should do in the context of this command being added?

Yes. I think for now it may be best to leave it as is and we can move it in a separate PR.

tleen:
> This would take away the either/or choice and let the command do the work rather than the user.

Yeah that is a good compromise between doing it like the other commands but allowing a separate file to be used as well:

```sh
$ GET https://raw.githubusercontent.com/exercism/lua/master/config.json > lua-nextercism-config.json
$ configlet tree lua-nextercism-config.json
```

cmd/visualize.go
tleen: Ok, it could involve some discussion. Once we get this in we can open up an issue about moving warnings or nextercism specific warnings into ui. :+1:

nywilken:
@tleen I'm a little late to the party, but great work. This is a solid PR. 

I have some thoughts on the command name and output, which I will leave inline. I'm mobile reviewing so please bare with me as my comments trickle in.

tleen:
@nywilken I appreciate any input on the PR. Thanks!

kytrinyx:
I like `tree` - short, sweet, familiar without being confusing due to punniness.

nywilken:
Regarding the command name, I actually think `list` gives us more flexibility for expansion and reads a little better the `tree` when read as `configlet list lua-config.json` vs `configlet tree lua-Config.json`

What do I mean by flexibility?
Well tree is a well known command that prints a hierarchy . But with a list command we could easy list all the exercises implemented by the track (flat view), and provide a `--tree` flag to get a tree format. Adding in other flags. I guess we could add a flatten to the tree command. 

The latter brings me to the output. I think we should be able to pipe the output in another program with easy. I will admit that I don't use the tree command all that much in my environment, but maybe a flag to show a flat list vs a whole tree would be nice.

tleen:
@nywilken For the piping, do you think the headers and stuff in the output is keeping the output from being well-pipeable or is the nature of the tree shape itself?

Do you envision the listing as a new line separated listing of exercise slugs, and `--tree` outputs as tree format?

kytrinyx:
> Regarding the command name, I actually think list gives us more flexibility

Yepp. That would let us add a --mode or --format flag or something to switch on which sort of list we want.

tleen:
So the consensus would be `list`, with the default output being a listing of all exercises, and a `--format` arg that would have `default` for the normal list and `tree` for the tree list?

nywilken:
> @nywilken For the piping, do you think the headers and stuff in the output is keeping the output from being well-pipeable or is the nature of the tree shape itself?

It really is more so the nature of the tree shape, with the delimiters. Headers seem to get separated but dashed lines so not bad to filter out. 

> Do you envision the listing as a new line separated listing of exercise slugs, and --tree outputs as tree format?

Yes, that's exactly it.

robphoenix:
I wonder if this is perhaps trying to pack too much into one command. Is there any reason we couldn't have both `list` and `tree` commands?

tleen:
It can be a tricky thing, designing command line apps with subcommands. You can look at it from the code perspective which would be "is there enough redundancy that I can fit this into one conceptual code unit with minimal additional code?" (Which in this case I think so.) But this is probably the wrong pattern of thought and we should look at what a user would expect. In this case it would be the user asking "how can I visualize my configuration file without having to get into the JSON?" When you look at it from this user perspective it may make more sense to have a single subcommand with the sole job of answering that question. Output formats just being a matter of immediate need.

I could see the usefulness of the `list` command as a generic place to put the visualizations. Even if for this pull we stick with only one (tree) or two (tree and newline-separated). I sort of took both approaches to visualizations in thinking this problem out through code: [single-purpose](https://github.com/tleen/exercism-config-tree) and [multiple-purpose](https://github.com/tleen/exercism-config-visualizations)!

If we go the single subcommand, multiple output formats route, would it make sense to not use `list` as the command name and use that as an output format? Perhaps the subcommand could be something as  simple as `output`:
```
configlet output . --fmt=list --with-difficulty
```

(I appreciate the opportunity to think out loud :mega: on this with you all. It is very helpful!)

kytrinyx:
> configlet output . --fmt=list --with-difficulty

I find this one difficult to mentally keep track of. Output _what_ exactly? And I think of `fmt` as re-formatting, and we already have a `fmt` command. So I worry that it will be like with the `tar` command where I can never remember which letters I need to use and the order of arguments.

So far I prefer tree, with the additional `--with-difficulty` flag. It's super explicit, and it's really about listing out the dependencies visually.

I'm also fine with `configlet list --tree [--with-difficulty]`, though I'm not certain that we actually need it.

tleen:
Ok, I think I'll go with `configlet list --tree [--with-difficulty]` and see how the maintainers like it.

Edit: I missed @Insti thumbs up there. Seems like its a higher majority for tree as it's own command so I'll stick with that for now. If we see some demand for  list we can talk about moving it in with tree or separating.

kytrinyx:
> If we see some demand for list we can talk about moving it in with tree or separating.

That sounds like a plan!

tleen:
I think this may have everything we need to add?

kytrinyx (approved):
🚢 This looks great!

kytrinyx:
@tleen Not to deliberately make more work for you, but would you be interested in being a co-maintainer on this repo?

robphoenix:
> @tleen Not to deliberately make more work for you, but would you be interested in being a co-maintainer on this repo?

I wholeheartedly endorse this. :heart:

tleen:
Sure, it would be my pleasure. I'm always looking for places to squeeze in more Go!

kytrinyx:
Yay. Done! And thank you ❤️

nywilken:
@tleen welcome!!!! I'm looking forward to working on more PRs with you.

tleen:
Same @nywilken glad to be able to help!

robphoenix:
@tleen woop woop! :tada: welcome :handshake:

PR #84, opened 2017-10-01
Remove coupling between description and metadata

### LINKED ISSUE #65: generate: <track>/<exercise>/.meta/description.md not replacing Spec.description
Not sure I understand the instructions entirely via the docs's [README readme](https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/readmes.md). 

Under "Description":
> Having said that, each track is free to write a more specific description if they want to, placing it either in the .meta/description.md, or directly into the README template.

I can't seem to get configlet to override the specification description using a `.meta/description.md` file inside of a track exercise. `.meta/hints.md` is included no problem. I am using configlet v3.4.0.

nywilken:
Currently, the loading of a custom exercise readme description (i.e `$exercise_path/.meta/description.md`)  requires that a copy of the `metadata.yml` file exist within `$exercise_path/.meta` . This change removes that coupling so that maintainers can override descriptions without the need of a duplicated `.metadata.yml` file. 

closes #65

### DIFF
--- a/track/problem_specification_test.go
+++ b/track/problem_specification_test.go
@@ -9,14 +9,16 @@ import (
 
 func TestNewProblemSpecification(t *testing.T) {
 	tests := []struct {
-		desc     string
-		slug     string
-		specPath string
-		expected ProblemSpecification
+		desc      string
+		slug      string
+		trackPath string
+		specPath  string
+		expected  ProblemSpecification
 	}{
 		{
-			desc: "shared spec if custom is missing",
-			slug: "one",
+			desc:      "shared spec if custom is missing",
+			slug:      "one",
+			trackPath: filepath.FromSlash("../fixtures/numbers"),
 			expected: ProblemSpecification{
 				Description: "This is one.\n",
 				Source:      "The internet.",
@@ -24,8 +26,9 @@ func TestNewProblemSpecification(t *testing.T) {
 			},
 		},
 		{
-			desc: "custom spec overrides shared",
-			slug: "two",
+			desc:      "custom spec overrides shared",
+			slug:      "two",
+			trackPath: filepath.FromSlash("../fixtures/numbers"),
 			expected: ProblemSpecification{
 				Description: "This is two, customized.\n",
 				Source:      "The web.",
@@ -33,29 +36,51 @@ func TestNewProblemSpecification(t *testing.T) {
 			},
 		},
 		{
-			desc:     "shared spec from alternate problem-specifications location",
-			slug:     "one",
-			specPath: filepath.FromSlash("../fixtures/alternate/problem-specifications"),
+			desc:      "shared spec from alternate problem-specifications location",
+			slug:      "one",
+			trackPath: filepath.FromSlash("../fixtures/numbers"),
+			specPath:  filepath.FromSlash("../fixtures/alternate/problem-specifications"),
 			expected: ProblemSpecification{
 				Description: "This is the alternate one.\n",
 				Source:      "The internet.",
 				SourceURL:   "http://example.com",
 			},
 		},
+		{
+			desc:      "custom spec metadata with shared description",
+			slug:      "metadata-example",
+			trackPath: filepath.FromSlash("../fixtures/granular-metadata-overrides"),
+			specPath:  filepath.FromSlash("../fixtures/granular-metadata-overrides/problem-specifications"),
+			expected: ProblemSpecification{
+				Description: "This is a shared description.\n",
+				Source:      "The web.",
+				SourceURL:   "",
+			},
+		},
+		{
+			desc:      "shared spec metadata with custom description",
+			slug:      "description-example",
+			trackPath: filepath.FromSlash("../fixtures/granular-metadata-overrides"),
+			specPath:  filepath.FromSlash("../fixtures/granular-metadata-overrides/problem-specifications"),
+			expected: ProblemSpecification{
+				Description: "This is a custom description.\n",
+				Source:      "The internet.",
+				SourceURL:   "http://example.com",
+			},
+		},
 	}
-
 	originalSpecPath := ProblemSpecificationsPath
 	defer func() { ProblemSpecificationsPath = originalSpecPath }()
 
 	for _, test := range tests {
 		ProblemSpecificationsPath = test.specPath
-		root := filepath.FromSlash("../fixtures")
-		spec, err := NewProblemSpecification(root, "numbers", test.slug)
+		root, trackID := filepath.Dir(test.trackPath), filepath.Base(test.trackPath)
+		spec, err := NewProblemSpecification(root, trackID, test.slug)
 		assert.NoError(t, err)
 
-		assert.Equal(t, test.expected.Description, spec.Description)
 		assert.Equal(t, test.expected.Source, spec.Source)
 		assert.Equal(t, test.expected.SourceURL, spec.SourceURL)
+		assert.Equal(t, test.expected.Description, spec.Description)
 	}
 }
 
--- a/fixtures/granular-metadata-overrides/exercises/description-example/.meta/description.md
+++ b/fixtures/granular-metadata-overrides/exercises/description-example/.meta/description.md
@@ -1,0 +1,1 @@
+This is a custom description.
--- a/fixtures/granular-metadata-overrides/exercises/metadata-example/.meta/metadata.yml
+++ b/fixtures/granular-metadata-overrides/exercises/metadata-example/.meta/metadata.yml
@@ -1,0 +1,2 @@
+---
+source: "The web."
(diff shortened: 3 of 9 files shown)


fixtures/elements/config.json
```diff
@@ -11,6 +11,12 @@
       "slug": "aluminum",
       "topics": [],
       "difficulty": 1
+    },
+    {
+      "uuid": "zzz",
```
kytrinyx: For fixtures I have the following general guideline:

* They should be nested within a directory that isolates them to a specific testing context.
* They should not be reused in different contexts
* The fixtures should be minimal, including only the features necessary to prove the point

If you look at the fixture tree, we're mostly following these principles. See "alternate", "format", "deprecated" (with different deprecations nested within them), and "broken-maintainers".


```
fixtures/
├── alternate
│   └── problem-specifications
│       └── exercises
│           └── one
│               ├── description.md
│               └── metadata.yml
├── broken-maintainers
│   ├── config
│   │   └── maintainers.json
│   └── config.json
├── broken.json
├── deprecated
│   ├── hints-both
│   │   ├── config.json
│   │   └── exercises
│   │       └── fake
│   │           └── HINTS.md
│   ├── hints-old
│   │   ├── config.json
│   │   └── exercises
│   │       └── fake
│   │           └── HINTS.md
│   ├── inserts-both
│   │   ├── config
│   │   │   └── exercise-readme-insert.md
│   │   ├── config.json
│   │   ├── docs
│   │   │   └── EXERCISE_README_INSERT.md
│   │   └── exercises
│   │       └── fake
│   └── inserts-old
│       ├── config.json
│       ├── docs
│       │   └── EXERCISE_README_INSERT.md
│       └── exercises
│           └── fake
├── elements
│   ├── config
│   │   └── maintainers.json
│   ├── config.json
│   └── exercises
│       └── aluminum
│           ├── example.ext
│           └── test.ext
├── empty.json
├── fake-exercise
│   ├── example.ext
│   ├── fake_test.ext
│   ├── specs
│   │   └── file.ext
│   └── subdir
│       └── solution.ext
├── format
│   ├── formatted
│   │   ├── config.json
│   │   └── maintainers.json
│   ├── malformed
│   │   ├── config.json
│   │   └── maintainers.json
│   ├── minimised
│   │   ├── config.json
│   │   └── maintainers.json
│   └── unformatted
│       └── config.json
├── numbers
│   ├── config
│   │   ├── exercise-readme-insert.md
│   │   ├── exercise_readme.go.tmpl
│   │   └── maintainers.json
│   ├── config.json
│   └── exercises
│       ├── one
│       │   ├── example.ext
│       │   └── test.ext
│       ├── three
│       │   └── tests
│       │       └── three.ext
│       ├── two
│       └── zero
└── problem-specifications
    └── exercises
        ├── fake
        │   ├── description.md
        │   └── metadata.yml
        ├── one
        │   ├── description.md
        │   └── metadata.yml
        └── two
            ├── description.md
            └── metadata.yml
```

I don't recall what I was thinking with "numbers" and "problem-specifications" nested at the top level. Probably that they were "happy path" situations. I'm not sure that's a good idea anymore, since I clearly can't remember 3 months later :(

For the functionality in this PR, we should have a nested fixture directory that is called something specific to the functionality, like `granular-metadata-overrides` and have whatever fixtures we need to fully specify the behavior within that directory.

track/problem_specification_test.go
```diff
@@ -44,6 +44,9 @@ func TestNewProblemSpecification(t *testing.T) {
 		},
 	}
 
+	saved := ProblemSpecificationsPath
```
kytrinyx: What does `saved` mean?

kytrinyx (changes requested):
The granular overrides look good—it looks like we basically have the same behavior as before, but on a file-by-file basis rather than making the assumption that we will load all the metadata from the same location.

I discovered when reading this that I have strong opinions about fixtures. Who knew? :-)

Fixtures are a huge source of pain over time, so getting them right as early as possible is an enormous win.

If it's not clear what I mean with my comments about fixtures, please discuss so we can hash it all out (and maybe we can document this afterwards; it feels important).

fixtures/elements/config.json
robphoenix: > I'm not sure that's a good idea anymore, since I clearly can't remember 3 months later :(

I know this feeling :smile: 

Out of interest, can I ask if this is a *Go thing*? I ask out of pure ignorance. For instance should the dir be called `testdata` instead of fixtures? (https://dave.cheney.net/2016/05/10/test-fixtures-in-go)

Would you split the directory up according to the command the fixture is being used to test? So have a sub-dir per command?

robphoenix (approved):
:+1:

fixtures/elements/config.json
nywilken: > For fixtures I have the following general guideline:
     - They should be nested within a directory that isolates them to a specific testing context.
     - They should not be reused in different contexts
     - The fixtures should be minimal, including only the features necessary to prove the point

Nice, this is helpful. I'm pretty sure I spent more time thinking how to organize the tests, and whether or not to create new fixtures than I did writing the fix. 

On the topic of fixtures, do you have a convention for naming? I struggle on just how granular to go. I try to lean towards shorter names, but I get how they may not provided enough context. Is it safe to say, if the fixture name properly explains its test context the length is irrelevant?

nywilken: > Out of interest, can I ask if this is a Go thing? I ask out of pure ignorance. For instance should the dir be called testdata instead of fixtures? (https://dave.cheney.net/2016/05/10/test-fixtures-in-go)

Great question, thanks for sharing the link.

kytrinyx: > Out of interest, can I ask if this is a Go thing?

Nope—this is purely me reacting to years of painful fixtures. I'd be happy to rename per Go conventions.

> Would you split the directory up according to the command the fixture is being used to test? So have a sub-dir per command?

Not really, unless it makes sense that way. Or rather—not _just_ sub commands. Take the "deprecations" for example. In other codebases I've worked in, we have one broad set of fixtures that have all the features that all the tests need. So you deprecate a field somewhere, and randomly (who knows which one) one of the fixtures will have the deprecated field, a different one will have the new field, and one will have both, so you can see which overrides which. Here I have a directory named `deprecation`, with one subdirectory for each specific deprecation. These fixtures don't need to have any other features, just the fields in question. And once we're ready to kill off the deprecated behavior, we can delete a subdirectory entirely.

When introducing an "alternate path" for problem-specifications, where you can pass it with a command-line switch, I added a specific subdirectory for it.

>  I try to lean towards shorter names, but I get how they may not provided enough context. Is it safe to say, if the fixture name properly explains its test context the length is irrelevant?

With fixtures (and tests in general) I tend to lean towards verbose and explicit. With a fixture I tend to want to know: under what circumstances would this need to be updated or deleted. So if we no longer wanted to support alternate paths for problem-specifications, or if there was a bug in how we were handling it, or we needed a third way to deal with it, it would all go in the same place.

Not sure if this helps—it's all just stuff I'm making up as we go along.

robphoenix: This is really helpful, thanks. I'm totally learning here, having ashamedly not implemented a lot of testing in my time :scream:, and as such don't have much to add. I like that the fixtures are grouped via behaviour than command, that makes sense. The more descriptive names make sense also.

I looked at changing `fixtures` to `testdata`, an opportunity to learn vim's [argdo](http://vimcasts.org/episodes/using-argdo-to-change-multiple-files/) :smile:, and it introduced a bug, as the exercise paths in the *fixtures/testdata* dir now contained the word `test` whether it had any tests in it or not, matching the test pattern [regex](https://github.com/exercism/configlet/blob/master/track/config.go#L33), thus implying there were tests present and passing linting where it should fail.

kytrinyx: Oh, man. I'm so often kicked in the gut by the assumptions that my tools make about what I want.

track/problem_specification_test.go
kytrinyx: Having seen a couple of other PRs go by, I realize that `saved` means `originalPath` - can we use originalSpecPath or origSpecPath here?

nywilken: You called this out in a different PR and I started moving away from saved. But I never answered you, my bad. I'll catch this one in the update 🙌🏽

nywilken:
@kytrinyx I made a few updates to reflect the comments about fixtures - the changes are pretty much, still, a work in progress. Once it's good to go I will ping the team for review.

nywilken:
@kytrinyx @robphoenix this is ready for review.

nywilken:
If there are no objections, I would like to merge this in and, possibly, start work for cutting a new release. Let me know 🙏

tleen (commented):
LGTM. Good fix here thanks!

robphoenix (approved):
:+1: go for it

tleen (approved):

nywilken:
Thanks!

PR #85, opened 2017-10-01
Add support for custom readme titles

### LINKED ISSUE #76: README generation: Sometimes we'd like different capitalisation of the title
My current example is the README of rna-transcription. The generator writes the title as "Rna Transcription", but it would be better if it were "RNA Transcription" instead. But how do we tell the generator that? I don't have a good solution.

Possible exercises that would benefit: DOT DSL, ETL, OCR Numbers, POV, RNA Transcription, SGF Parsing

Related that should probably go in a different issue: punctuation such as Pascal's Triangle, Diffie-Hellman.

nywilken:
closes #76

### DIFF
--- a/cmd/lint_test.go
+++ b/cmd/lint_test.go
@@ -15,6 +15,12 @@ import (
 )
 
 func TestLintTrack(t *testing.T) {
+	originalNoHTTP := noHTTP
+	noHTTP = true
+	defer func() {
+		noHTTP = originalNoHTTP
+	}()
+
 	originalOut := ui.Out
 	originalErrOut := ui.ErrOut
 	ui.Out = ioutil.Discard
@@ -23,6 +29,7 @@ func TestLintTrack(t *testing.T) {
 		ui.Out = originalOut
 		ui.ErrOut = originalErrOut
 	}()
+
 	lintTests := []struct {
 		desc     string
 		path     string
@@ -240,9 +247,9 @@ func TestDuplicateTrackUUID(t *testing.T) {
 	ts := httptest.NewServer(fakeEndpoint)
 	defer ts.Close()
 
-	saved := UUIDValidationURL
+	originalUUIDValidationURL := UUIDValidationURL
 	UUIDValidationURL = ts.URL
-	defer func() { UUIDValidationURL = saved }()
+	defer func() { UUIDValidationURL = originalUUIDValidationURL }()
 
 	expected := []string{"ccc"}
 	track := track.Track{
--- a/fixtures/titled-exercises/config.json
+++ b/fixtures/titled-exercises/config.json
@@ -1,0 +1,12 @@
+{
+  "slug": "titled-exercises",
+  "language": "titled-exercises",
+  "exercises": [
+    {
+      "slug": "slug-as-title"
+    },
+    {
+      "slug": "explicit-title"
+    }
+  ]
+}
--- a/track/exercise_readme_test.go
+++ b/track/exercise_readme_test.go
@@ -26,18 +26,39 @@ func TestNewExerciseReadme(t *testing.T) {
 }
 
 func TestGenerateExerciseReadme(t *testing.T) {
-	readme := ExerciseReadme{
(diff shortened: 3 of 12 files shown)


cmd/lint_test.go
```diff
@@ -13,6 +13,10 @@ import (
 )
 
 func TestLintTrack(t *testing.T) {
+	saved := disableHTTPChecks
```
kytrinyx: What does `saved` mean? Can we name it specific to what we're saving?

track/exercise_readme_test.go
```diff
@@ -26,18 +26,39 @@ func TestNewExerciseReadme(t *testing.T) {
...
-		Spec: &ProblemSpecification{
-			Slug:        "hello-kitty",
-			Description: "The description.\n",
+	readmeTests := []struct {
+		description string
+		Spec        *ProblemSpecification
+		expected    string
+	}{
+		{
+			description: "readme with title cased slug for .Spec.Name.",
```
kytrinyx: Maybe "readme with title inferred from slug"?

track/problem_specification.go
```diff
@@ -75,6 +82,10 @@ func (spec *ProblemSpecification) Credits() string {
 	return fmt.Sprintf("%s [%s](%s)", spec.Source, spec.SourceURL, spec.SourceURL)
 }
 
+func (spec *ProblemSpecification) titleCaseSlug() string {
```
kytrinyx: When naming things that return a value, I try to name them based on _what they are_ rather than _what they do_.

I don't know what the best name is here. _InferredTitle_? _NamifiedSlug_? _TitleCasedSlug_?

kytrinyx (changes requested):
This looks good. I have a couple of small comments, but nothing major.

robphoenix (approved):
LGTM :+1:

track/problem_specification.go
nywilken: Good advice. I'll update. 

> I don't know what the best name is here. InferredTitle? NamifiedSlug? TitleCasedSlug?

I'm inclined to use `titleCasedSlug` as ProblemSpecification types should use the Name method for getting the readme title. Do you think it the method should be exported?

cmd/lint_test.go
nywilken: I see how saved can be confusing.

nywilken:
@kytrinyx @robphoenix thanks for the first round of reviews.

track/problem_specification.go
kytrinyx: I don't think it needs to be exported, since we'll be using Name (or whatever it is) to be either the explicit title or the inferred one.

kytrinyx (approved):

nywilken:
@kytrinyx @robphoenix I know I got approval on this PR already, but I did make some changes to the tests and fixtures that I would like to get your approval on before I merge. 🙇

fixtures/titled-exercises/config.json
```diff
@@ -0,0 +1,20 @@
...
+      "uuid": "yyy",
+      "slug": "slug-as-title",
+      "topics": [],
+      "difficulty": 1
+    },
+    {
+      "uuid": "zzz",
+      "slug": "explicit-title",
+      "topics": [],
+      "difficulty": 4
```
kytrinyx: Is the difficulty relevant to the titles? If not I would set this to 1 (since 1 seems to be the default). Or even better—if the test doesn't blow up I would delete all but the slug field.

kytrinyx (approved):
I have a tiny quibble with the fixture, but as soon as that is addressed I would say 🚢 -- this looks great! 🌷

fixtures/titled-exercises/config.json
nywilken: Good call!

nywilken:
@petertseng support for custom titles is now part of [configlet](https://github.com/exercism/configlet/releases/tag/v3.6.0). Please let me us know if you run into any issues when generating new READMEs.

PR #88, opened 2017-10-03
redirect output from ui package in lint test file

robphoenix:
This suppresses stdout & stderr output from the lint test when running the test file.

### DIFF
--- a/cmd/lint_test.go
+++ b/cmd/lint_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"path/filepath"
@@ -9,10 +10,19 @@ import (
 	"testing"
 
 	"github.com/exercism/configlet/track"
+	"github.com/exercism/configlet/ui"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestLintTrack(t *testing.T) {
+	originalOut := ui.Out
+	originalErrOut := ui.ErrOut
+	ui.Out = ioutil.Discard
+	ui.ErrOut = ioutil.Discard
+	defer func() {
+		ui.Out = originalOut
+		ui.ErrOut = originalErrOut
+	}()
 	lintTests := []struct {
 		desc     string
 		path     string


cmd/lint_test.go
```diff
@@ -9,10 +10,20 @@ import (
 	"testing"
 
 	"github.com/exercism/configlet/track"
+	"github.com/exercism/configlet/ui"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestLintTrack(t *testing.T) {
+	var bufOut, bufErrOut bytes.Buffer
+	savedOut := ui.Out
```
nywilken: Im thinking we should agree on some standard. I know saved as a variable is used in TGPL book, which was my reason for using it, but what do you think about origOut vs savedOut?

I went back and forth within my PR and settled on orig as the prefix because I found it gave the variables a bit more context. I could be over thinking it 😐 

But I'm open to hear thoughts for the sake of consistency and less thinking later on.

cmd/lint_test.go
```diff
@@ -9,10 +10,20 @@ import (
 	"testing"
 
 	"github.com/exercism/configlet/track"
+	"github.com/exercism/configlet/ui"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestLintTrack(t *testing.T) {
+	var bufOut, bufErrOut bytes.Buffer
```
nywilken: Maybe  `ioutil.Discard` would be a better option as we don't actually care about the output for these tests.

nywilken (commented):
@robphoenix the changes look good. I have a small suggestion pertaining to the output overrides, and a question about naming.

tleen (commented):
This is a good fix. Since the buffers aren't really doing anything with the text perhaps it is a good place for a /dev/null style writer: [ioutil.Discard](https://golang.org/pkg/io/ioutil/#pkg-variables). It makes the purpose of the code immediately obvious.

Edit: @nywilken beat me to it above. :+1:

cmd/lint_test.go
kytrinyx: I like _original_ in terms of context. (For what it's worth. 👋 sorry my presence has been a bit spotty lately)

robphoenix: @kytrinyx :wave: no need to apologise, ever, I can't even begin to imagine what it must be like carrying the weight of an OSS project like this. I stress out sometimes just maintaining the Go track. You are very much an important and positive force here, but please do, always, take all the time you need to be in a good place.

nywilken: > For what it's worth.

@kytrinyx your feedback and guidance is always appreciated. I can't thank you enough for the guidance, and time, you have given us. Like @robphoenix said, there is no need to apologize, ever.

kytrinyx (approved):
🌷 ship it!!!

Issue #80, opened 2017-09-23
lint: duplicateTrackUUID check fails if the repository name is not correct


icyrockcom:
I have a fork of the purescript track, but I still have the old repository name:

https://github.com/icyrockcom/xpurescript

Looks like the track ID is determined based on the repository folder name:

https://github.com/exercism/configlet/blob/master/track/track.go#L27

In a sample build in my personal travis build setup:

https://travis-ci.org/icyrockcom/xpurescript/builds/278841675

the linter fails with:

```
...
$ bin/configlet lint .
-> The following UUID was found in multiple Exercism tracks. Each exercise UUID must be unique across tracks.
c41a2961-f7ca-4eaf-83cf-48762fc39c06
...
```

since it thinks the track is called `xpurescript`, not `purescript`.

A few ways to resolve this with the current setup come to mind:

- Rename the repository, which in my case is fine, but other people may not be able to do that. For example, someone can be working on a fork of https://github.com/purescript/purescript/ at the same time
- Use subfolders (e.g. a repository called exercism, then subfolders for each of the tracks as e.g. lpil has done with his https://github.com/lpil/exercism)
- Edit the travis configuration to rename the folder after checkout (e.g. such as recommended here https://stackoverflow.com/a/16480058/438544)

In order to avoid falling back to these workarounds, it might be better to store the track ID in config.json and read it from there, with a fall-back to the folder name.

kytrinyx:
This is a good point, thank you.

> Rename the repository, which in my case is fine, but other people may not be able to do that.

Yeah, we've run into that a couple of times with language contributors who also contribute to Exercism tracks.

> In order to avoid falling back to these workarounds, it might be better to store the track ID in config.json and read it from there, with a fall-back to the folder name.

I removed the track ID from the config.json since it wasn't being used anywhere on the site or by the tools, but this is a valid use case, and I think that having the track ID in the config is the cleanest solution.

I can script this up to submit the change to all the tracks.

icyrockcom:
Ah, I did not know it was there in the first place :) OK, cool, it'll be easy for me to test it out afterwards.

kytrinyx:
@robphoenix @nywilken there are a couple of ways of doing this. We could either provide the track ID in all the config.json files, or we could update the commands to take a `--track-id` flag (optional, default to the directory name), and update the tracks to use that.

Thoughts? Preferences?

robphoenix:
[...]

PR #90, opened 2017-10-04
Add --track-id flag to lint command

robphoenix:
This is a tool specific solution to #80

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -6,17 +6,22 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/exercism/configlet/track"
 	"github.com/exercism/configlet/ui"
 	"github.com/spf13/cobra"
 )
 
-// UUIDValidationURL is the endpoint to Exercism's UUID validation service.
-var UUIDValidationURL = "http://exercism.io/api/v1/uuids"
-
-// noHTTP flag indicates if HTTP-based lint checks have been disabled at runtime.
-var noHTTP bool
+var (
+	// UUIDValidationURL is the endpoint to Exercism's UUID validation service.
+	UUIDValidationURL = "http://exercism.io/api/v1/uuids"
+	// noHTTP flag indicates if HTTP-based lint checks have been disabled at runtime.
+	noHTTP bool
+	// trackID flag allows the user to specify the ID of the track,
+	// for example if it is different to the local directory name
+	trackID string
+)
 
 // lintCmd defines the lint command.
 var lintCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


track/track.go
```diff
@@ -58,3 +58,11 @@ func New(path string) (Track, error) {
 	}
 	return track, nil
 }
+
+// RenameID renames the track id,
+// it only does this with non-empty strings
+func (t *Track) RenameID(name string) {
```
nywilken: I get the idea, but this code can live directly within lintTrack. No need for an exported function on track.

track/track_test.go
```diff
@@ -60,3 +60,32 @@ func TestTrackID(t *testing.T) {
 		os.Chdir(cwd)
 	}
 }
+
+func TestTrackRenameID(t *testing.T) {
```
nywilken: I think we can do away with this test. See my comment above about dropping the function RenameID

nywilken (changes requested):
I left a comment on the issue, pertaining to this PR. I also left a few comments, if we end up going this route. 

Thanks for pushing this fix along. Great stuff!

track/track.go
robphoenix: Ahh you're so right, I totally over complicated that. :roll_eyes: Thanks!

cmd/lint.go
```diff
@@ -318,4 +337,5 @@ func duplicateTrackUUID(t track.Track) []string {
 func init() {
 	RootCmd.AddCommand(lintCmd)
 	lintCmd.Flags().BoolVar(&disableHTTPChecks, "no-http", false, "Disable remote HTTP-based linting.")
+	lintCmd.Flags().StringVar(&trackID, "track-id", "", "Specify track name.")
```
nywilken: I think the messaging needs to be more descriptive. Specifically, it should explain why a user would want to specify a track name and what it defaults to. Maybe something like `Override the default track-id with 'name' (defaults to directory name)`

nywilken (approved):
@robphoenix this looks good - just one minor nit pick before we merge.

cmd/lint.go
kytrinyx: I'd prefer _Specify the track ID_ rather than _Specify the track name_, since name is ambiguous. The track name is the name of the language, which is different from the ID.

cmd/lint.go
```diff
@@ -6,17 +6,22 @@ import (
...
-var UUIDValidationURL = "http://exercism.io/api/v1/uuids"
-
-// disableHTTPChecks flag skips HTTP based checks when passed.
-var disableHTTPChecks bool
+var (
+	// UUIDValidationURL is the endpoint to Exercism's UUID validation service.
+	UUIDValidationURL = "http://exercism.io/api/v1/uuids"
+	// disableHTTPChecks flag skips HTTP based checks when passed.
+	disableHTTPChecks bool
+	// trackID flag allows the user to specify the name of the track,
```
kytrinyx: Ditto here for s/name/ID/

kytrinyx (approved):
Other than the bit about track name vs track ID, this looks great. Good to merge as soon as that is fixed.

PR #93, opened 2017-10-10
Rename fixtures to match testing context

nywilken:
I've reorganized a number of the fixtures, which I added previously, to match the guidelines listed below.

- They should be nested within a directory that isolates them to a specific testing context.
- They should not be reused in different contexts
- The fixtures should be minimal, including only the features necessary to prove the point


cmd/lint_test.go
```diff
@@ -47,7 +47,7 @@ func TestLintTrack(t *testing.T) {
 		},
 		{
 			desc:     "should not fail when given a track with all of its bits in place.",
-			path:     "../fixtures/elements",
+			path:     "../fixtures/lint/valid-track",
```
kytrinyx, line 50: This makes me so happy 😂 I clearly have a warped sense of priorities in life 🌻

kytrinyx (approved):
This looks great. Thanks so much for taking the time to sort through this.

cmd/lint_test.go
nywilken: 😆

PR #95, opened 2017-10-12
Add generate command info to README

### LINKED ISSUE #91: Update our own README
[yo dawg](http://knowyourmeme.com/memes/xzibit-yo-dawg) I heard you like READMEs....

We should probably update our own README to include info on generating READMEs with the configlet tool. I have an outline in the Go track [README](https://github.com/exercism/go#exercism-go-style) that I generated with configlet's README `generate`. We could start with that? 

Happy to do this once I get `tree` done.

tleen:
Add information about generating track READMEs from the `generate`
command. Not the semantics of the command (available from --help)
rather the information about template files, data overrides and their
locations.

Resolves #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,50 @@ Inspired by Go's [`gofmt`](https://blog.golang.org/go-fmt-your-code) tool, confi
     * Exercises will have their list of topics sorted alphabetically.
     * Topics names will be normalised to be lowercase and contain underscores in place of spaces.
 
+## Generate
+
+The configlet `generate` command may be used to generate README's for all exercises on the track, as a whole or individually. The overall purpose of this command is to utilize common [`problem-specifications`](https://github.com/exercism/problem-specifications) data in generating a *template-based* README for each exercise, while at the same time permitting overrides of this data on a track exercise basis.
+
+Using this command for README generation allows for some conveniences:
+
+1. The use of a track specific README template file.
+1. The automatic inclusion of information from the `problem-specifications` repository.
+1. Allow overrides of that `problem-specification` data on a per-exercise basis.
+1. The ability to insert language specific exercise information ("hints") to an exercise README.
+1. Full customization of a README template per exercise (if necessary).
+
+`generate` looks for specific files in the track root's `config/` directory and in each exercise's `.meta/` directory. It will then use these files, if present, and the `problem-specifications` data to generate an exercise's README. 
+
+Much of the utility of this command comes from the ability to *locally override* README templates and exercise information.
+
+### The README Template
+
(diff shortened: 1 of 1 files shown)


ilya-khadykin:
Wow, thanks a lot for this, @tleen!

README.md
```diff
@@ -26,6 +26,50 @@ Inspired by Go's [`gofmt`](https://blog.golang.org/go-fmt-your-code) tool, confi
...
+
+#### .Spec.Description
+
+This variable is sourced from an exercise's [`description.md`](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/description.md) file in the `problem-specifications` repo. You may override this variable's contents for an exercise by adding a `.meta/description.md` file in that track exercise's directory.
+
+#### .Spec.Credits
+
+The credits are a description of the source of an exercise with an optional hyperlink to that source. This information originates from the [`metadata.yml`](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/metadata.yml) located in the exercise's `problem-specifications` entry. You may override this information for an exercise by adding a `.meta/metadata.yml` file in that track exercise's directory.
+
+#### .Spec.Name
```
nywilken, line 63: There are two alternative versions of `.Spec.Name` (i.e `.Spec.MixedCaseName` and `.Spec.SnakeCaseName`) that can be used within README templates for tracks wishing to include filename generation within their READMEs. Here's an example from the [Groovy track](https://github.com/exercism/groovy/blob/master/config/exercise_readme.go.tmpl#L13)

Should they be included?

nywilken (commented):
@tleen this looks great, thank you for updating the README. I left one comment for review.

README.md
tleen: Totally, nice catch! I think I should lump them in with .Spec.Name?

nywilken: > I think I should lump them in with .Spec.Name?

That makes sense, especially since the other two methods are alternate versions of name. It may help to include the Groovy code snippet as an example or some other example use case.

ferhatelmas (approved):
LGTM

README.md
tleen: Will do. Thanks for the pointer on Groovy. I am trying to spread around the linked examples to different repos so the docs have that "useful in all tracks" feel and no particular track looks like the One True Way.

tleen:
Added a link to [grip](https://github.com/joeyespo/grip) in the README where it deals with READMEs. I have found it valuable as it lets you locally render and preview the [GFM](https://github.github.com/gfm/), helps catch mistakes.

nywilken (approved):
This looks great - detailed and easy to follow. Feel free to merge when ready.

tleen:
Thanks :smile:

Issue #102, opened 2017-10-23
configlet tree: crashes on "null"


rpottsoh:
Update: `tree` is crashing when it comes across `"null"` when it is expecting `null`.

`configlet tree .` is crashing when I run it against the CSharp track.  It is working fine against Delphi and Ruby.

Here is the output I am receiving when I run it against the CSharp track:
```text
C:\projects.git\Exercism\csharp>configlet tree .
C#
==
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x2c pc=0x667f17]

goroutine 1 [running]:
github.com/exercism/configlet/cmd.treeTrack(0x11c40740, 0x1, 0x8a1b20, 0x11c34690)
        /home/wilken/Development/golang/src/github.com/exercism/configlet/cmd/tree.go:235 +0x717
github.com/exercism/configlet/cmd.runTree(0x8a1b20, 0x11c34690, 0x1, 0x1)
        /home/wilken/Development/golang/src/github.com/exercism/configlet/cmd/tree.go:110 +0x50
github.com/exercism/configlet/vendor/github.com/spf13/cobra.(*Command).execute(0x8a1b20, 0x11c34680, 0x1, 0x1, 0x8a1b20, 0x11c34680)
        /home/wilken/Development/golang/src/github.com/exercism/configlet/vendor/github.com/spf13/cobra/command.go:653 +0x1dc
github.com/exercism/configlet/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x8a1a00, 0x11c48000, 0x11c48000, 0x403f73)
        /home/wilken/Development/golang/src/github.com/exercism/configlet/vendor/github.com/spf13/cobra/command.go:728 +0x264
github.com/exercism/configlet/vendor/github.com/spf13/cobra.(*Command).Execute(0x8a1a00, 0x1, 0x11c3a6c0)
        /home/wilken/Development/golang/src/github.com/exercism/configlet/vendor/github.com/spf13/cobra/command.go:687 +0x21
github.com/exercism/configlet/cmd.Execute()
        /home/wilken/Development/golang/src/github.com/exercism/configlet/cmd/root.go:42 +0x27
main.main()
        /home/wilken/Development/golang/src/github.com/exercism/configlet/main.go:6 +0x17
```
Configlet lint is not reporting any problems with CSharp.  I am using the 32 bit Windows version of Configlet.  Let me know if you need anymore information.

[...]

PR #103, opened 2017-10-24
tree: Fix invalid unlocked_by reference panic 

tleen:
See #102


rpottsoh:
Thanks @tleen for making the necessary changes.

cmd/tree.go
```diff
@@ -230,8 +230,17 @@ func treeTrack(configFilepath string) error {
 			continue
 		}
 
+		parent, unlockExists := slugToExercise[e.UnlockedBy]
```
nywilken, line 233: Do we need to run this type of exists check in other areas of the code where we expect to get a reference to an exercise?

ferhatelmas (changes requested):
Seems test is missing `gofmt`. Otherwise good.

nywilken (approved):
@tleen this looks good to me. I am a little concerned that the warning may get missed, especially for tracks with a large set of exercises.

I left a question about your change. But this is otherwise good to merge, when you are ready.

cmd/tree_example_test.go
```diff
@@ -84,3 +84,27 @@ func ExampleTreeWarnings() {
...
+	// ----
+	// ├─ one
+	// │  └─ five
+	// │
+	// └─ two
+	//
+	// bonus
+	// -----
+	// seven
+}
```
tleen: @ferhatelmas is this the line that makes you think it did not pass through `gofmt`?

tleen: @ferhatelmas good eye on that, there are a few places it was wrong. Dunno why it didn't run on save as usual.

cmd/tree.go
tleen: I don't believe so, not in `tree` anyway. The rest are referenced via the main slug which is the authoritative map key for an exercise. 

Unsure about other commands. I think once `lint` checks for this stuff it is ok for other commands to assume that config won't have bad slugs.

tleen: @nywilken the warning could be missed, but not sure how to make it stand out more?

cmd/tree.go
```diff
@@ -230,8 +230,17 @@ func treeTrack(configFilepath string) error {
...
+		// An unlocked_by slug that does not exist and is referenced can crash
+		// the program, see #102. If an non-existant slug exists issue warning.
+		if !unlockExists{
+			printConfigurationWarning(
+				fmt.Sprintf("Exercise \"%s\" has an invalid unlocked_by slug: \"%s\"", e.Slug, e.UnlockedBy));
```
ferhatelmas: `s/%s/%q/` seems better for formatting.

cmd/tree_example_test.go
```diff
@@ -84,3 +84,27 @@ func ExampleTreeWarnings() {
 	// one
 	// two
 }
+
+func ExampleInvalidUnlockedBy(){
```
ferhatelmas: Sorry about nitpick. There should be a space after parens.

tleen: No you're right, something has gone wonky with my `gofmt`. It did everything but that.

cmd/tree.go
tleen: :100: better. Great call.

cmd/tree.go
```diff
@@ -230,8 +230,17 @@ func treeTrack(configFilepath string) error {
 			continue
 		}
 
+		parent, unlockExists := slugToExercise[e.UnlockedBy]
+		// An unlocked_by slug that does not exist and is referenced can crash
+		// the program, see #102. If an non-existant slug exists issue warning.
```
ferhatelmas: Also, `a non-existent slug` ? https://en.wiktionary.org/wiki/non-existant

tleen: heh, a non-existant word.

cmd/tree.go
```diff
@@ -232,7 +232,7 @@ func treeTrack(configFilepath string) error {
 
 		parent, unlockExists := slugToExercise[e.UnlockedBy]
 		// An unlocked_by slug that does not exist and is referenced can crash
-		// the program, see #102. If an non-existant slug exists issue warning.
+		// the program, see #102. If an non-existent slug exists issue warning.
```
ferhatelmas: `a non-...` :wink:

ferhatelmas (approved):

PR #109, opened 2017-11-09
lint: refactoring for higher abstractions

ferhatelmas:
Implementation of checkers in lint command started to be [repetitive](https://github.com/exercism/configlet/pull/98/files#r145857260). This tries to create an abstraction to make adding a new checker easier.

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -52,9 +52,7 @@ func lintExampleText() string {
 func runLint(cmd *cobra.Command, args []string) {
 	var hasErrors bool
 	for _, arg := range args {
-		if failed := lintTrack(arg); failed {
-			hasErrors = true
-		}
+		hasErrors = lintTrack(arg) || hasErrors
 	}
 	if hasErrors {
 		os.Exit(1)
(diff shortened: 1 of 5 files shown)


petertseng:
I subscribe, so that if this is merged before #98 I get notified that I should rebase #98.

ferhatelmas:
@petertseng sorry, I had thought you're already checking all PRs in this repo :wink: but I should have notified you anyway.

cmd/lint.go
```diff
@@ -52,9 +52,7 @@ func lintExampleText() string {
 func runLint(cmd *cobra.Command, args []string) {
 	var hasErrors bool
 	for _, arg := range args {
-		if failed := lintTrack(arg); failed {
-			hasErrors = true
-		}
+		hasErrors = lintTrack(arg) || hasErrors
```
kytrinyx, line 55: This is a nice improvement: it manages to be both more concise and more readable at the same time.

cmd/lint.go
```diff
@@ -127,185 +125,111 @@ func lintTrack(path string) bool {
...
-		metadata[exercise.Slug] = false
-	}
-	// Don't report missing implementations on foregone exercises.
-	for _, slug := range t.Config.ForegoneSlugs {
-		metadata[slug] = true
-	}
-	for _, exercise := range t.Exercises {
-		metadata[exercise.Slug] = true
-	}
+type folder struct{ add, remove []string }
```
kytrinyx, line 134: This type is named `folder`, which can be confusing in a few ways. It's named as though it's an interface (a generic thing that has a `Fold` method), but this is not an interface, it's a struct—a thing—so my brain reads it as the thing "a folder", which is a synonym for a directory. But it's not that, either. It's a _thing that performs fold operations on stuff that it holds on to_. This is made even more confusing, because we actually do have `Fold` methods implemented further down.

It took me several rounds of reading the entire change in the PR before I finally understood what it is.

This confusion between _what a thing is_, and _what a thing does_ is further reflected in the names of the collections that it holds. The names `add` and `remove` are imperative verbs, but they're naming a _collection_ of things.

In general, when naming collections, I try to name them with the plural of a noun. The reason I use a plural is because it really helps me think straight when I'm reading code to be able to iterate through a collection and read "for each **monkey** in **monkeys**" rather than "for each thing in add".

```diff
+ for monkey := range monkeys {
- for e := range add {
  // ...
}
```

A lot of people reach for collective nouns, but I find that as confusing as iterating over `add`. E.g.:

```diff
+ // a flock of birds, a school of fish, a dazzle of zebras... 
+ for monkey := range tribe {
- for e := range add {
  // ...
}
```

At a high level, I think that you're right about the duplication that you've identified, and also that the concept of folding is probably a good way to address it, at least to a certain degree.

The problem with the name _fold_ is that it can mean a lot of things. In this PR there are at least 3 different operations all named _fold_, and it's not clear what the important differences between them are without reading the implementations.

When we have multiple things that are similar and related, I find that it is helpful to choose names that help compare and contrast. If they're similar but unrelated, I want to names that help make the distance between them even further. The names are the clues that help fit the pieces together into a coherent whole.

_Folding_ is a fundamental concept in computer science, though. It's a low-level operation performed on data types. Using the name _fold_ makes sense at the low-level parts of the program for generic abstraction over a data type, that you would only see if reading the details of implementation.

Here, though, the name is problematic. We're at the top level of the program, and I would prefer that the names of things and actions tell the story of the domain: linting Exercism tracks. We're _figuring out_ or _ensuring_ or _detecting_ specific things about tracks and exercises. These higher level ideas might use some of the lower level computational concepts, but it shouldn't communicate _in terms of_ those things.

cmd/lint.go
```diff
@@ -127,185 +125,111 @@ func lintTrack(path string) bool {
...
+	}
+	return m
 }
 
-func missingMetadata(t track.Track) []string {
-	implementations := map[string]bool{}
-	for _, exercise := range t.Exercises {
-		implementations[exercise.Slug] = false
-	}
+func asFolder(add, remove []string) folder { return folder{add: add, remove: remove} }
```
kytrinyx, line 146: I struggled for a bit to figure out which abstraction mapped to the abbreviation `as` before realizing that this is a factory, and that it's creating a `folder` using the two slices.

The amount of code is approximately equal in the two expressions:

```
// 1.
folder{add: []string{...}, remove: []string{...}}

// 2.
asFolder([]string{...}, []string{...})
```

If we're going to name something, I want to be certain that the name communicates an idea that is more meaningful than the implementation it would be replacing. For example in the below, "highlighting" is a domain concept about _why_ we're toggling the background to be yellow. Both are true, one idea is an implementation detail that might change.

```
func highlight(p string) string {
    return toggle(p, Node::BG, Color::YELLOW)
}
```

In the case of `asFolder` I don't think the name adds any conceptual clarity.

cmd/lint.go
```diff
@@ -127,185 +125,111 @@ func lintTrack(path string) bool {
...
+func reversed(f folder) folder             { f.add, f.remove = f.remove, f.add; return f }
```
kytrinyx, line 147: The name `reverse` or `reversed` doesn't say what it reverses. Until I read the implementation, I figured it reversed the order of the strings in the slices (though I wasn't sure why we would need that).

cmd/lint.go
```diff
@@ -127,185 +125,111 @@ func lintTrack(path string) bool {
...
+	for _, f := range fs {
+		m = f.fold(m)
 	}
+	return m
+}
 
-	for _, exercise := range t.Config.Exercises {
-		implementations[exercise.Slug] = true
-	}
+func foldSlugsByExistence(fs ...folder) []string {
```
kytrinyx, line 157: This name is about the computational operation, "fold this". While I completely understand the operation, I find it very hard to tie it back to the _why_ of the program. Why are we folding? What exists? Is it important that it exist? Or should it not exist?

cmd/lint.go
```diff
@@ -127,185 +125,111 @@ func lintTrack(path string) bool {
...
+	return foldSlugsByCount(0, fs...)
+}
 
-	slugs := []string{}
-	for slug, ok := range implementations {
-		if !ok {
+func foldSlugsByCount(threshold int, fs ...folder) []string {
```
kytrinyx, line 161: Here, too, I can read the code to understand what it's doing, but the name itself doesn't tell me. It says we're folding, and it says that a count is involved, but I have to read the implementation in order to understand what the operation means or is trying to achieve.

What would really help me here is if the names helped me understand what we're actually looking for or trying to detect... in the domain of linting an Exercism track.

cmd/lint.go
```diff
@@ -127,185 +125,111 @@ func lintTrack(path string) bool {
...
-		uuids = append(uuids, exercise.UUID)
-	}
-
 	payload := struct {
 		TrackID string   `json:"track_id"`
 		UUIDs   []string `json:"uuids"`
 	}{
 		TrackID: t.ID,
-		UUIDs:   uuids,
+		UUIDs:   t.Config.ExerciseUUIDs(false),
```
kytrinyx, line 232: What does `withEmpty` mean? What would it mean when it's not empty, and where do we use it?

In general I try to avoid boolean parameters for functions and methods, because they invariably mean that I have to jump to the implementation to either figure out what it means, or to remind myself which option it refers to. In this case I read the implementation but still didn't understand what it meant.

If passing the option is necessary, then I would prefer to have a custom type with explanatory name.

track/exercise.go
```diff
@@ -15,6 +15,21 @@ type Exercise struct {
 	TestSuitePath string
 }
 
+// Exercises is a slice of Exercise to define functions
+type Exercises []Exercise
```
kytrinyx, line 19: I like that the name of the collection is the plural of the thing that it contains. We're back to the monkey/monkeys thing again.

cmd/lint.go
```diff
@@ -127,185 +125,111 @@ func lintTrack(path string) bool {
...
-			slugs = append(slugs, slug)
-		}
+func (f folder) fold(m map[string]int) map[string]int {
+	for _, slug := range f.add {
+		m[slug]++
 	}
-	return slugs
+	for _, slug := range f.remove {
+		delete(m, slug)
+	}
```
kytrinyx, line 142: The names `add` and `remove` sound like they should be symmetrical, because they are opposite operations. However with `add` we are incrementing, but with `remove` we are deleting, which is confusing to me.

track/exercise.go
```diff
@@ -15,6 +15,21 @@ type Exercise struct {
...
+
+// Fold iterates over slice, runs given function and collects slugs
+func (es Exercises) Fold(isValid func(Exercise) bool) (valid []string, invalid []string) {
```
kytrinyx, line 22: The function name `Fold` suggests that it's a generic folding function, but the parameter and return values suggest that this is a much more specific folding function that has to do with validation.

However, because of the way it's named, this would not be apparent in the code where it is used, only when you read the implementation:

````
coll1, coll2 := es.Fold(someFn)
```

I think it would be helpful to have a name that helps express this idea that we're partitioning the exercises based on validity.

track/exercise_metadata.go
```diff
@@ -12,3 +12,21 @@ type ExerciseMetadata struct {
 	IsCore       bool   `json:"core"`
 	IsDeprecated bool   `json:"deprecated"`
 }
+
+// HasUUID checks if UUID is defined
+func (e ExerciseMetadata) HasUUID() bool { return e.UUID != "" }
+
+// ExerciseMetadataList is a slice of ExerciseMetadata to define functions
+type ExerciseMetadataList []ExerciseMetadata
```
kytrinyx, line 20: I'm now coming to realize how problematic the name `Metadata` is, because we tend to use it as an uncountable thing (a pile of "stuff", not individual items that can be enumerated). It makes it tricky to try to name a collection, since there's no good plural for it. I want to avoid using the name `List` for a number of reasons. First it's an overloaded term. We use it casually in some contexts (e.g. "shopping list", "a list of my favorite things"), but it also has a very specific meaning in computer science. Here it feels like you ended up with the word 'List' because you need to name the collection, but there's no good name for the collection of things it will contain. If that makes sense.

kytrinyx (changes requested):
Overall I agree that there's duplication, and that it's probably duplication that we can (and should) address.

I'm quite uncertain about the particulars, and I'm not yet sure whether I think that the abstraction isn't quite right yet, or if the abstraction is fine, but that the names just need work.

I really appreciate that you're taking the time to explore this. 🌻 The fact that I'm so torn about the details is a good sign that there's something important for me to learn in here!

nywilken:
@ferhatelmas I see there hasn't been any movement on this PR since November. Do you think you will have time to address @kytrinyx comments? If not we can close for now and revisit a restructure at a later time. Let me know. Thanks!

ferhatelmas:
@nywilken It went out of scope. @kytrinyx gave wonderful review. I will address them and come back with a newer one. Thanks for taking care of repo.

PR #110, opened 2017-11-12
track: check error on readme generation

ferhatelmas:
Error checking was missed.


kytrinyx (approved):
🌷 thanks for catching this.

PR #115, opened 2018-02-04
Add filter for unsupported exercise directories

### LINKED ISSUE #108: configlet lint: ignore git-ignored files?
The current version of `configlet` reports the following lint errors on the Java track:

```
$ configlet lint .
-> An implementation for '.idea' was found, but config.json does not reference this exercise.
-> An implementation for '.gradle' was found, but config.json does not reference this exercise.
-> The implementation for '.idea' is missing an example solution.
-> The implementation for '.gradle' is missing an example solution.
-> The implementation for '.gradle' is missing a test suite.
```

The directories found here are local-only (ignored by git). Is there a way to have `configlet lint` operate only on non-git-ignored files?

nywilken:
There is currently an issue where hidden directories (those beginning with a `.` or `_`) are being treated as exercise slugs if they reside under `track-root/exercises/...`. This change introduces a regular expression that skips any directory under `track-root/exercises` that begins with either a `.` or `_`. 

closes #108

### DIFF
--- a/fixtures/unsupported-exercise-names/config.json
+++ b/fixtures/unsupported-exercise-names/config.json
@@ -1,0 +1,13 @@
+{
+  "slug": "unsupported-exercise-names",
+  "repository": "https://github.com/exercism/unsupported-exercise-names",
+  "active": true,
+  "solution_pattern": "[Ee]xample|\\.meta/solutions/[^/]*\\.ext",
+  "exercises": [
+    {
+      "slug": "some-valid-exercise",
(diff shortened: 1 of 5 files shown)


kytrinyx (approved):
🚢 This is lovely! The fixtures are nice and well-contained, the test is easy to understand, and the code intervention is minimal and super clean. Love it!

nywilken:
Sweet! Thanks for the review and feedback.

PR #118, opened 2018-02-25
Add Upgrade command

### LINKED ISSUE #112: Add upgrade command
We have an upgrade command in the Exercism CLI, which would be handy to have in Configlet.

I kind of don't want to extract that logic into a separate package and create a whole new dependency that we need to track, so I'm tempted to suggest that we copy it in.

On the other hand, copy/paste might be a terrible idea.

What do you all think?

nywilken:
This change introduces a command for upgrading to the latest available version of Configlet. The upgrade command, is identical to the upgrade command used in the Exercism CLI. In fact, the underlying pkg `github.com/exercism/cli/cli` is being pulled into Configlet to download and upgrade the Configlet binary in-place. This change resolves #112 

**Note**: The pkg version of `github.com/exercism/cli/cli` is currently pinned to the Nextercism branch.

**Follow-Up**: The existing pkg structure of `github.com/exercism/cli/cli` seems a bit strange so I think a follow up, in the near future, might be to restructure the pkg or even break it out of `exercism/cli` into its own pkg for upgrading GitHub released binaries.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -16,14 +16,14 @@
 [[projects]]
   branch = "nextercism"
   name = "github.com/exercism/cli"
-  packages = ["config","debug"]
-  revision = "9c2c5857da4af84af2973a04f68402116e1b5f43"
+  packages = ["cli","config","debug"]
+  revision = "63dae2f74c28b9bcfaf1920dd57603c3b8971b4b"
 
 [[projects]]
   name = "github.com/fsnotify/fsnotify"
   packages = ["."]
-  revision = "629574ca2a5df945712d3079857300b5e4da0236"
-  version = "v1.4.2"
+  revision = "c2828203cd70a50dcccfb2761f8b1f8ceef9a8e9"
+  version = "v1.4.7"
 
 [[projects]]
   name = "github.com/google/uuid"
@@ -41,7 +41,7 @@
   branch = "master"
   name = "github.com/hashicorp/go-multierror"
   packages = ["."]
-  revision = "83588e72410abfbe4df460eeb6f30841ae47d4c4"
+  revision = "b7773ae218740a7be65057fc60b366a49b538a44"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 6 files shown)


Gopkg.lock
```diff
@@ -1,12 +1,30 @@
...
 [[projects]]
   name = "github.com/davecgh/go-spew"
   packages = ["spew"]
   revision = "346938d642f2ec3594ed81d874461961cd0faa76"
   version = "v1.1.0"
 
+[[projects]]
+  branch = "nextercism"
+  name = "github.com/exercism/cli"
+  packages = ["cli","config","debug"]
```
kytrinyx: I see what you mean by strange package structure. In the future, let's see if we can do this with a single import (it would be nice if we could set the values we need on the upgrade package rather than pulling in a dependency on config and debug).

kytrinyx (approved):
I'm not super happy with having to pull in three packages from exercism/cli—and that we also have to hard-code the release url twice.

I'd love to see if we can get that cleaned up before too long.

I think your idea of a completely separate package (something like `exercism/ghrelease` or `exercism/ghpkg` maybe?).

nywilken:
@kytrinyx thanks for the review.

> I'm not super happy with having to pull in three packages from exercism/cli—and that we also have to hard-code the release url twice.

The release URL being in there twice is a complete oversight on my part. I'll cleaned things up and pushed a new change. 

Regarding the three packages, I was thinking about using the `ignored` property in GoDep since we don't actually use `config` or `debug` to reduce the dependency graph, but wasn't sure if it would confuse things. I updated the `Gopkg.*` files in my latest push. Let me know what you think.

> I think your idea of a completely separate package (something like exercism/ghrelease or exercism/ghpkg maybe?).

I like `exercism/ghrelease`. Breaking it out could be a good experiment and could be a first step in building a pkg for managing binaries through GitHub releases.

kytrinyx (approved):
This is much nicer. Thanks for cleaning that up!

If you have the time to take a stab at a ghrelease package, I think that would be a neat experiment.

nywilken:
> If you have the time to take a stab at a ghrelease package, I think that would be a neat experiment.

@kytrinyx I saw this comment but never got around to responding. I will definitely give this a shot and keep you posted on the progress. BTW `github/hub` is sick. Thanks for introducing me to it.

kytrinyx:
> BTW github/hub is sick.

Yes! It's so good. The GitHub API is great and all, but the CLI does so much to make it easy to work against the API.

> I will definitely give this a shot and keep you posted on the progress.

Sweet. No hurry, obviously :) I've been struggling just to keep up with emails, much less actually make progress on stuff.

PR #119, opened 2018-02-26
Lint: Checks for Invalid Unlocked By & Locked Core Exercises

### LINKED ISSUE #39: Update lint command: verify that core exercises are never unlocked by other exercises
Per @jonmcalder's (fabulous) comment here: https://github.com/exercism/discussions/issues/175#issuecomment-319067625 we should probably do some linting for the track progression logic.

If an exercise is `core` it should not be `unlocked_by` another exercise.

### LINKED ISSUE #61: lint: verify that exercises are only ever unlocked by core exercises.
Per https://github.com/exercism/v2-feedback/issues/61#issuecomment-324130763

If an exercise is `unlocked_by` another, we should make sure that the exercise it depends on is a core exercise.

tuxagon:
I've implemented 2 checks for linting detailed in #39 
1. Any core exercise should not be unlocked by another exercise
2. All unlocked by exercises should be valid exercises

It's been awhile since I've done some go, so I hope I wasn't too rusty on this one. Let me know what needs done to improve it!

Edit: fixes #39, fixes #61


cmd/lint.go
```diff
@@ -357,6 +365,40 @@ func duplicateTrackUUID(t track.Track) []string {
...
+func unlockedByValidExercise(t track.Track) []string {
+	slugs := []string{}
+	valid := map[string]bool{}
+
+	for _, exercise := range t.Config.Exercises {
+		valid[exercise.Slug] = true
+	}
+
+	for _, exercise := range t.Config.Exercises {
+		unlockedBy := strings.Trim(exercise.UnlockedBy, " ")
```
nywilken: What do you think if we dropped this transformation and processed the `exercise.UnlockedBy` as is? 

Basically causing the lint command to fail if an entry contains extra spaces. Which may encourage maintainers of the `config.json` files to clean up the config files so that UnlockedBy values are exact Slug matches.

cmd/lint.go
```diff
@@ -357,6 +365,40 @@ func duplicateTrackUUID(t track.Track) []string {
 	return []string{}
 }
 
+func lockedCoreViolation(t track.Track) []string {
+	slugs := []string{}
+	for _, exercise := range t.Config.Exercises {
+		if exercise.IsCore && len(strings.Trim(exercise.UnlockedBy, " ")) > 0 {
```
nywilken: I believe the default value for `exercise.UnlockedBy` will be an empty string if it is not defined for an exercise within `config.json`. Which means we might be able to drop the call to `strings.Trim` to remove whitespace and change line 371 to check if an exercise IsCore and has a non-empty UnlockedBy value. 

Which raises the question if a config.json file with an entry like the following is valid (e.g a space for the unlocked_by feild):
```javascript
{
   "Slug": "bar",
    "core": true,
    "unlocked_by": " "
}
```

cmd/lint.go
```diff
@@ -357,6 +365,40 @@ func duplicateTrackUUID(t track.Track) []string {
...
+	valid := map[string]bool{}
+
+	for _, exercise := range t.Config.Exercises {
+		valid[exercise.Slug] = true
+	}
+
+	for _, exercise := range t.Config.Exercises {
+		unlockedBy := strings.Trim(exercise.UnlockedBy, " ")
+
+		if len(unlockedBy) == 0 {
```
nywilken: Should we skip core exercises in this check since they are being checked elsewhere?

nywilken (commented):
@tuxagon thanks for submitting this PR. I took an initial look and the overall flow looks good. I left a few questions and suggestions.

cmd/lint.go
tuxagon: Sure thing! That does make sense.

cmd/lint.go
tuxagon: The default I found most common is actually `null`, but that translates to an empty string. Either way, it does make sense to be less permissive and require a cleaner `unlocked_by` value.

cmd/lint.go
tuxagon: Hmm. I think it actually might be superfluous to check for core because the idea is that I would assume the core exercises are configured properly here. Based on that, the updated conditional, `if len(exercise.UnlockedBy) == 0 || exercise.IsCore` would be a little redundant. Thoughts?

cmd/lint_test.go
```diff
@@ -292,3 +292,81 @@ func TestDuplicateTrackUUID(t *testing.T) {
...
+		},
+	}
+
+	slugs := lockedCoreViolation(track)
+
+	assert.Equal(t, 1, len(slugs))
+	assert.Equal(t, "apple", slugs[0])
+}
+
+func TestAcceptableUnlockBy(t *testing.T) {
```
nywilken: Since you are using table tests you can consolidate the two test functions  for lockedCoreViolation into one test function with varying test cases.

cmd/lint_test.go
```diff
@@ -292,3 +292,81 @@ func TestDuplicateTrackUUID(t *testing.T) {
...
+			},
+		},
+	}
+
+	slugs := lockedCoreViolation(track)
+
+	assert.Equal(t, 0, len(slugs))
+}
+
+func TestValidUnlockedByExercises(t *testing.T) {
```
nywilken: I recommend renaming the function so that it reflects the functionality being tested as opposed to th exact use case under test. In other words call it something like `TestUnlockedByViolations‘ so that we can easily add new test cases of different inputs without the function name losing its meaning.

fixtures/invalid-unlocked-by/config.json
```diff
@@ -0,0 +1,25 @@
+{
```
nywilken: Were these committed by accident ? I don't believe they are being used nor are they needed by your tests. With that said if you plan to add some additional tests that leverage the fixtures then I would recommend you remove everything you don't need (i.e exercise files, or non required metadata that is relevant to the test)

nywilken (commented):
@tuxagon thanks for the updates this is looking good - just a few minor suggestions before we merge.

cmd/lint.go
```diff
@@ -357,6 +365,38 @@ func duplicateTrackUUID(t track.Track) []string {
...
+	for _, exercise := range t.Config.Exercises {
+		if exercise.IsCore && len(exercise.UnlockedBy) > 0 {
+			slugs = append(slugs, exercise.Slug)
+		}
+	}
+
+	return slugs
+}
+
+func unlockedByValidExercise(t track.Track) []string {
```
nywilken, line 379: @kytrinyx can I get your help in clarifying what is a valid value for unlocked by. I just saw PR #120 for unlocked_by only being core exercises. I validated in the docs https://github.com/exercism/docs/blob/master/language-tracks/configuration/exercises.md but I found a couple of exercises where unlocked bys are not core. 

If only core exercises can be used as valid unlocked_by values we can refactor this method a bit. 

Thanks 🙏

cmd/lint_test.go
tuxagon: I'm not sure what you mean by table tests, but I could collapse the 2 tests using a structure like 

```go
trackTests := []struct {
	desc               string
	expectedViolations int
	expectedSlug       string
	config             track.Config
}{}

// . . .

for _, tt := range trackTests {
	track := track.Track{Config: tt.config}
	slugs := lockedCoreViolation(track)

	assert.Equal(t, tt.expectedViolations, len(slugs), tt.desc)

	if len(slugs) > 0 {
		assert.Equal(t, tt.expectedSlug, slugs[0], tt.desc)
	}
}
```

Is this what you mean? Or is this an acceptable resolution?

fixtures/invalid-unlocked-by/config.json
tuxagon: I wasn't sure about them because I noticed they existed for other lints, so I included them. I can remove them; no big deal. :+1:

cmd/lint_test.go
nywilken: Sorry for not providing a link around table tests (https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go). But what you put in your comment is exactly what I mean. Using a structure to define multiple inputs, and expected outputs that could be iterated over to test a function with different cases.

tuxagon: Ok. I'll make the necessary changes. Thanks.

tuxagon: I updated, but GitHub didn't apparently notice the change, so I think you might have to manually resolve or request a further change.

fixtures/locked-core-exercise/config.json
```diff
@@ -0,0 +1,27 @@
+{
```
nywilken: The fixture directory for `locked-core-exercise` can be deleted.

cmd/lint.go
```diff
@@ -121,6 +121,14 @@ func lintTrack(path string) bool {
 			check: duplicateTrackUUID,
 			msg:   "The following UUID was found in multiple Exercism tracks. Each exercise UUID must be unique across tracks.\n%v",
 		},
+		{
+			check: lockedCoreViolation,
+			msg:   "The exercise '%v' is marked as core and unlocked by another exercise. A core exercise should not be unlocked by another.",
+		},
+		{
+			check: unlockedByValidExercise,
+			msg:   "The exercise '%v' is being unlocked by a non-existent test.",
```
nywilken: ~~non-existent test~~ a non-existent exercise. 

If it turns out that only core exercises can unlock others then we would want to change the text to "a non-core exercise".

cmd/lint.go
nywilken: @kytrinyx @tuxagon I opened a discussion at https://github.com/exercism/discussions/issues/211 to help get some clarity around this question. If it turns out that only valid core exercises can unlock others, we can change line 384 to set true for core exercises or even add a guard clause to skip anything that is not core.

nywilken (commented):
@tuxagon the updated test file looks good - great work! I left a few minor nitpicks on the PR and left a comment on the idea of a valid unlocking exercise being a core exercise. 

Until we here back, would you mind pulling down the latest master and rebasing your branch so that you have the latest changes.

cmd/lint.go
nywilken: @tuxagon it turns out that a valid exercise must be an existing core exercise. Please update your logic, and test cases, to validate that exercises are being unlocked by valid core exercises. Thank you again for the contribution and patience with the review process.

kytrinyx: Just to confirm: yepp, the current requirements of v2 are that any non-core exercises can only be unlocked by core exercises.

tuxagon: Roger that :+1:

tuxagon:
I think I've updated according to all the suggestions. Did I miss anything I still need to do?

tuxagon:
Fixes #39, #61 for GitHub to auto-close. @cmccandless, thanks for your sibling contribution. I just wanted to recognize it here because I feel a little bad.

cmccandless:
@tuxagon 
> Fixes #39, #61

This needs to be in the PR description to allow auto-close.

No worries, grafting the functionality added by #120 into this PR results in a much cleaner change.

nywilken:
@tuxagon @cmccandless thanks again for the contributions.

tuxagon:
@nywilken Thanks for the time reviewing!

PR #125, opened 2018-03-28
Gopkg: correct uuid dependency

ferhatelmas:


Gopkg.toml
```diff
@@ -24,8 +24,8 @@
 ignored = ["github.com/exercism/cli/config","github.com/exercism/cli/debug" ]
```
ferhatelmas, line 24: @nywilken why are they ignored ? 

Also, I think we should install dependencies via `dep` in travis if we do vendoring.

Gopkg.toml
```diff
@@ -24,8 +24,8 @@
...
 
 [[constraint]]
-  branch = "master"
-  name = "github.com/mattetti/uuid"
+  name = "github.com/google/uuid"
```
nywilken, line 27: Good catch - I noticed this after updating to 1.10 but never got around to creating the PR. Thanks.

Gopkg.toml
nywilken: The config and debug pkgs pull in a lot of unnecessary dependencies so they are being ignored to remove the cruft. The cli  pkg under exercism/cli doesn’t actually use them (well maybe debug), but because they are in the same pkg tree they get pulled in. 

I am experimenting with the creation of a GitHub Release pkg which would replace the cli pkg.

nywilken (approved):
Good to merge.

PR #126, opened 2018-03-28
all: linting

ferhatelmas:
* drop unnecessary error check
* check shallowed error
* drop unnecessary `return`
* simplify regex literal


nywilken (approved):
This looks good to me. Thanks for the cleanup.

PR #127, opened 2018-04-22
Update configlet fmt to apply a definite ordering

coriolinus:
The ordering used is specified in
https://github.com/exercism/meta/issues/95#issuecomment-341991512

This simplifies the track-specific decisions as to whether or not
to use configlet fmt, because it now outputs config.json in an
ordering friendlier to humans than alphabetical.

For an example of a decision so simplified, see exercism/rust#516.

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -37,18 +37,24 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
 // formatter applies additional formatting to unmarshalled JSON files.
 type formatter func(m map[string]interface{})
 
+// orderer applies an ordering to unmarshalled JSON files
+type orderer func(map[string]interface{}) OrderedMap
+
 func runFmt(cmd *cobra.Command, args []string) {
 	path := args[0]
 	var fs = []struct {
 		path string
 		formatter
+		orderer
 	}{
 		{
 			filepath.Join(path, "config.json"),
 			formatTopics,
+			orderConfig,
 		},
 		{
 			filepath.Join(path, "config", "maintainers.json"),
+			nil,
 			nil,
 		},
 	}
@@ -60,7 +66,7 @@ func runFmt(cmd *cobra.Command, args []string) {
 			ui.PrintError("path not found:", f.path)
 			os.Exit(1)
 		}
-		diff, formatted, err := formatFile(f.path, f.formatter)
+		diff, formatted, err := formatFile(f.path, f.formatter, f.orderer)
 		if err != nil {
 			ui.PrintError(err.Error())
 			continue
@@ -84,7 +90,7 @@ func runFmt(cmd *cobra.Command, args []string) {
 	}
 }
 
-func formatFile(path string, format formatter) (diff string, formatted []byte, err error) {
+func formatFile(path string, format formatter, order orderer) (diff string, formatted []byte, err error) {
 	f, err := os.Open(path)
 	if err != nil {
 		return diff, formatted, err
@@ -99,12 +105,19 @@ func formatFile(path string, format formatter) (diff string, formatted []byte, e
 		format(m)
 	}
 
+	var om interface{}
+	if order == nil {
+		om = m
+	} else {
+		om = order(m)
+	}
+
 	original, err := ioutil.ReadFile(path)
 	if err != nil {
 		return diff, formatted, err
 	}
 
-	formatted, err = json.MarshalIndent(&m, "", "  ")
+	formatted, err = json.MarshalIndent(&om, "", "  ")
 	if err != nil {
 		return diff, formatted, err
 	}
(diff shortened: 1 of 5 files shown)


nywilken:
@coriolinus thanks for submitting this change! I'll work on reviewing it and getting back to you with some feedback.

cmd/fmt.go
```diff
@@ -37,19 +37,25 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
 // formatter applies additional formatting to unmarshalled JSON files.
 type formatter func(m map[string]interface{})
 
+// orderer applies an ordering to unmarshalled JSON files
+type orderer func(map[string]interface{}) OrderedMap
```
nywilken, line 41: Have you considered using the the `Track` and `Maintainer` types for achieving the expected ordering? 

I see the approach you took and it make sense, but I believe that we can roll this formatting into the respective types to avoid duplicating fields and adding complexity in maintaining two types. 

Might even want to consider an interface type, which both `Track` and `Maintainer` could implement to keep format signatures the same.

cmd/fmt.go
```diff
@@ -161,6 +174,45 @@ func normaliseTopic(t string) string {
 	return s
 }
 
+// orderConfig applies an ordering to config.json
+//
+// This ordering was specified in
+// https://github.com/exercism/meta/issues/95#issuecomment-341991512
+func orderConfig(m map[string]interface{}) OrderedMap {
```
nywilken, line 181: This is a pretty impressive function; nice work. If we leverage the underlying types `Track` we can order the fields to confirm to the spec and remove the need for the custom ordering.

nywilken:
@coriolinus thanks again for submitting this change. I’m looking forward to getting the ordering in. I left a few comments around the solution. I think we can take a different approach seeing as we are now enforcing order.

 By using the respective structs we can get the order we want without having to manipulate maps. I will add more comments, but want to get you some feedback to start.

coriolinus:
Hi, sorry this has taken so long.

In short: no, I didn't consider editing the core types. My goal while writing this was to make the smallest set of changes necessary to accomplish the task. As you can tell by the delay in response, I just don't have the necessary free time to devote to grokking the full system. I therefore picked the fastest path toward victory. 

I can see the appeal of editing the core types instead of imposing an external ordering here. However, the more I think about it, the less I'm convinced that that approach is actually better. One of two things must be true: either we would need to enumerate the order of the fields in the type definition, in which case we haven't gained anything over this implementation, or we'd need to hard-code the order of the fields in each type's `MarshalJSON` method, in which case extensibility becomes a problem if the structs ever need to grow. In either case, writing two custom `MarshalJSON` methods would necessarily involve writing more code than writing one generic one.

Or perhaps I'm wrong, and there's a simple and elegant way to have the types generate an internal ordering instead of using an `OrderedMap` to impose an external one. However, if such a thing exists, I don't see it.

kytrinyx:
As far as I know, the JSON package orders keys when marshalling—struct keys are marshalled in the order defined in the struct.

kytrinyx:
I'd love to get this merged in the next day or two, as I'm about to participate in ROSSConf (Friday/Saturday), where we'll be doing a 1-day hackday/workshop for people to contribute to various open source projects, and it would be great to get the config stuff straightened out before the workshop.

@coriolinus if you don't have time to make changes in the next couple of days that's fine, just please let us know :)

@nywilken if @coriolinus doesn't have time to make changes, how would you feel about merging this and doing a subsequent refactor, or making some tweaks on this PR before merging?

coriolinus:
I'm sorry, I'm pretty tied up in work stuff right now, so I can't promise
any real time devoted to this in the next few days. I'd also love to see
this merged soon; it unblocks the Rust track from upgrading its CI
guarantees.

I certainly don't mind if this code is all torn out in a refactor; what I
care about is that configlet actually orders the keys, not the precise
method by which it accomplishes that.

nywilken:
@kytrinyx @coriolinus 

> if @coriolinus doesn't have time to make changes, how would you feel about merging this and doing a subsequent refactor, or making some tweaks on this PR before merging?

This sounds good to me. I think we should merge the PR in its current state as it does fix the intended problem and refactor post release. This way we can unblock the Rust track and give adequate time to update configs prior to ROSSConf.

kytrinyx:
awesome, agreed!

kytrinyx:
I'll cut a release in the next day or so.

nywilken:
:laughing: I was a few seconds to slow on hitting merge, thanks! I have bandwidth to cut the release today if it's just a straight forward release.

kytrinyx:
> I have bandwidth to cut the release today if it's just a straight forward release.

Oh, that would be fantastic! Yeah, there's nothing special about the release, just the regular cut.

nywilken:
@kytrinyx @coriolinus I went ahead and cut the release for v3.9.0.  You should be able to upgrade by running `configlet upgrade`

PR #131, opened 2018-05-27
Distinguish between undefined and empty for topics

### LINKED ISSUE #129: fmt: Converts empty list to null value
Running `configlet fmt <file>` converts an empty list to a null value, which is semantically different.

See https://github.com/exercism/erlang/pull/244#issuecomment-388557811.

kytrinyx:
When the maintainers have not yet considered topics for an exercise, the value should be null.
If they have decided that there are no relevant topics for an exercise, the value should be an
empty array.

Fixes #129 

FYI @laurafeier

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -142,11 +142,14 @@ func formatTopics(m map[string]interface{}) {
 		if !ok {
 			continue
 		}
+
 		topics, ok := exercise["topics"].([]interface{})
 		if !ok {
+			// Topics are null.
 			continue
 		}
-		var sorted []string
+		// Ensure that topics are an empty list rather than null.
+		sorted := make([]string, 0)
 		for _, t := range topics {
 			topic, ok := t.(string)
 			if !ok {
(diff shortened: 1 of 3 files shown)


cmd/fmt.go
```diff
@@ -141,11 +141,14 @@ func formatTopics(m map[string]interface{}) {
...
 		}
+
 		topics, ok := exercise["topics"].([]interface{})
 		if !ok {
+			// Topics are null.
 			continue
 		}
-		var sorted []string
+		// Ensure that topics are an empty list rather than null.
+		sorted := make([]string, 0)
```
nywilken: Since we know the maximum size of sorted at this time we could set the capacity to len(topics) reducing the number of times we allocate the underlying array ```sorted := make([]string, 0, len(topics))```

nywilken (approved):
This is solid, nicely done. I left one comment around setting the initial capacity for the sorted topics slice, but that should no prevent a merge. So I am approving the PR as it stands.

cmd/fmt.go
kytrinyx: Ah, yepp, that would be much better. I'll change that and then get this merged. Thank you!

PR #132, opened 2018-05-27
Rework the fmt command and tests

kytrinyx:
I want to refactor the fmt command to rely on the default ordering based on Struct fields rather than the OrderedMap implementation, but I didn't trust the tests enough to make the change.

This makes a series of smaller tweaks that provide a better entry point for the tests.

The new tests run the full command passing a "root" directory. The code now allows for passing a separate "out" directory where the reformatted configs will be written. By default the out directory is the same as the in directory, but the tests override this with a temporary directory. We then read and compare the input file and the output file.

Perhaps the most important change, though, is that the tests now run whatever formatting and ordering that the `runFmt` command uses, instead of manually passing those to the formatFile function, which means that the implementation could have been quite badly wrong without our knowing it.

I will do the larger refactoring to remove the ordered map implementation in a subsequent PR.


kytrinyx:
I've reworked this to go on top of the changes that went into master this morning, and it's ready for review.

kytrinyx:
I realized that we need to fix the ordering of the maintainer config; this is done in #133. Then the final refactoring that gets rid of the ordered map is in #134.

cmd/fmt.go
```diff
@@ -40,65 +46,66 @@ type formatter func(m map[string]interface{})
...
 
 	if changes != "" {
 		ui.Print("changes made to:\n", changes)
 	}
+	return nil
 }
 
-func formatFile(path string, format formatter, order orderer) (diff string, formatted []byte, err error) {
-	f, err := os.Open(path)
+func formatFile(inPath, outPath string, format formatter, order orderer) (diff string, err error) {
```
nywilken: The named returns are not needed here if we move the declaration of `diff` into the function, but otherwise the simplified signature looks good.

cmd/fmt.go
```diff
@@ -112,24 +119,26 @@ func formatFile(path string, format formatter, order orderer) (diff string, form
...
 	}
 
-	diff, err = difflib.GetUnifiedDiffString(
-		difflib.UnifiedDiff{
-			A: difflib.SplitLines(string(original)),
-			B: difflib.SplitLines(string(formatted)),
-		})
+	src := difflib.SplitLines(strings.TrimSuffix(string(original), "\n"))
+	dst := difflib.SplitLines(string(formatted))
+	diff, err = difflib.GetUnifiedDiffString(difflib.UnifiedDiff{A: src, B: dst})
```
nywilken: Clean :+1:

cmd/fmt_test.go
```diff
@@ -2,93 +2,68 @@ package cmd
...
 
-var (
-	expectedConfig      string
-	expectedMaintainers string
-)
-
-func TestMain(m *testing.M) {
-	cfg, err := ioutil.ReadFile(filepath.FromSlash("../fixtures/format/formatted/config.json"))
+func TestFmtCommand(t *testing.T) {
+	cfgTrack, err := ioutil.ReadFile(filepath.FromSlash("../fixtures/format/formatted/config.json"))
```
nywilken: `trackCfg` and `maintainersCfg` may read a little better. I was expecting a directory when I first read the variable name.

nywilken (commented):
This looks good. This refactor simplifies things quite a bit and brings clarity around the tests. I left a few comments around variable declarations and naming. But otherwise good to go.

cmd/fmt.go
kytrinyx: I mostly included the named returns for documentation purposes. When looking at just the signature of the method it's not obvious what the string that is being returned is (the formatted file contents?). I'll go ahead and remove them and if it gets confusing later we can put them back in.

cmd/fmt_test.go
kytrinyx: Yeah, agreed. Good call!

PR #134, opened 2018-05-27
Refactor the fmt command to rely on default marshaling of structs

kytrinyx:
Instead of using the OrderedMap which was introduced so that we could use alphabetic ordering of fields, this changes the fmt command to rely on the order that the fields are defined in the structs.


kytrinyx:
This rebases the branch onto the most recent master. It should be good to review properly now.

nywilken:
@kytrinyx thanks for rebasing. I'm looking through the changes and should have some feedback for you in the am.

track/config.go
```diff
@@ -64,3 +74,34 @@ func NewConfig(path string) (Config, error) {
 	}
 	return c, nil
 }
+
+// Read loads a config from file given the path to the file.
+func (cfg *Config) Read(path string) error {
```
nywilken: The method name `Read` seems like it should be something like `func (cfg *Config) LoadFromFile(path string) error` or maybe even `func NewConfigFromFile(path string) (Config, error)` so that it is a tad more explicit that it is replacing the contents of the existing Config. Thoughts?

track/maintainer_config.go
```diff
@@ -45,3 +46,21 @@ func NewMaintainerConfig(path string) (MaintainerConfig, error) {
 	}
 	return mc, nil
 }
+
+// Read loads a config from file given the path to the file.
+func (mCfg *MaintainerConfig) Read(path string) error {
```
nywilken: See my comment on `Config.Read` in `config.go`

track/config_serializer.go
```diff
@@ -0,0 +1,6 @@
+package track
+
+type ConfigSerializer interface {
```
nywilken: What do you think about defining this interface in the `fmt` command? Seeing as it is only used there.

track/config_serializer.go
```diff
@@ -0,0 +1,6 @@
...
+	Read(string) error
+	ToJSON() ([]byte, error)
```
nywilken: The ToJSON signature has a striking resemblance to the Marshaler interface `MarshalJSON ([]byte, error)`. However, I can't seem to get a working solution to recommend a change here. So I will leave it as a point of reference for something to think about in the future https://blog.gopheracademy.com/advent-2016/advanced-encoding-decoding/

cmd/fmt.go
```diff
@@ -40,253 +37,75 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
...
-type formatter func(m map[string]interface{})
-
-// orderer applies an ordering to unmarshalled JSON files
-type orderer func(map[string]interface{}) OrderedMap
-
 func runFmt(inDir, outDir string, verbose bool) error {
+	// This is for the tests.
+	// We want to be able to test the written output
+	// without clobbering the fixtures.
+	os.Mkdir(filepath.Join(outDir, "config"), os.ModePerm)
```
nywilken: Should we add a clean up for this? Or a flag that we can toggle only when in test mode.

nywilken (commented):
Nicely done - breaking up the commit the way you did made reviewing easier. I have a few nitpicks, but nothing crazy. 

I ran the fmt command a few times and it appears to be working as expected. The messaging when given an invalid config is not useful `invalid character ...`, maybe we can add a more helpful error.

track/config.go
kytrinyx: Agreed, that's much better.

track/config_serializer.go
kytrinyx: Yes, I like that better. Thanks!

cmd/fmt.go
kytrinyx: Maybe we can check if `outDir` is different from `inDir` and only create the directory if they're different. Or only create it if it doesn't exist.

```
$ go doc os.Mkdir
func Mkdir(name string, perm FileMode) error
    Mkdir creates a new directory with the specified name and permission bits
    (before umask). If there is an error, it will be of type *PathError.
```

My assumption was that Mkdir will not do anything if the directory exists, but the documentation doesn't specify. I need to check if it changes permissions in the case where it does exist.

kytrinyx:
@nywilken thanks so much for the feedback! I'll clean these things up and ping you for a final look.

kytrinyx:
Ok, I think I've addressed the feedback. Would you mind taking another look @nywilken?

cmd/fmt.go
```diff
@@ -40,253 +37,74 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
...
 
-// formatter applies additional formatting to unmarshalled JSON files.
-type formatter func(m map[string]interface{})
-
-// orderer applies an ordering to unmarshalled JSON files
-type orderer func(map[string]interface{}) OrderedMap
-
 func runFmt(inDir, outDir string, verbose bool) error {
+	if _, err := os.Stat(filepath.Join(outDir, "config")); os.IsNotExist(err) {
+		os.Mkdir(filepath.Join(outDir, "config"), os.ModePerm)
```
nywilken, line 43: :+1:

cmd/fmt.go
```diff
@@ -40,253 +37,74 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
...
-	// It will go away in a subsequent refactoring.
-	os.Mkdir(filepath.Join(outDir, "config"), os.ModePerm)
-
 	var changes string
 
 	for _, f := range fs {
-		diff, err := formatFile(f.inPath, f.outPath, f.formatter, f.orderer)
+		diff, err := formatFile(f.cfg, f.inPath, f.outPath)
 		if err != nil {
 			return err
```
nywilken: I recommend using multierror pkg we are using for the Generate command here. I found a bug when running  the `fmt` command on a track that is missing `maintainers.json`. Seems like we should lint for it, but that's another issue. 

If you take any track that has an improperly formatted config and remove it maintainers config running `configlet fmt <path/to/track>` will exit with an error stating that the maintainers.json was not found, which would imply that the command failed. But if you run a `git status` it will show that the config.json was actually formatted. 

I recommend that we display errors encountered along with any changes made. Something like the following:
```
> configlet fmt .                         
-> changes made to:
 config.json

-> 1 error occurred:
* open config/maintainers.json: no such file or directory
```
Sample Patch
```
diff --git a/cmd/fmt.go b/cmd/fmt.go
index 1ddea74..0dda129 100644
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/exercism/configlet/track"
 	"github.com/exercism/configlet/ui"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pmezard/go-difflib/difflib"
 	"github.com/spf13/cobra"
 )
@@ -61,10 +62,12 @@ func runFmt(inDir, outDir string, verbose bool) error {
 
 	var changes string
 
+	errs := &multierror.Error{}
 	for _, f := range fs {
 		diff, err := formatFile(f.cfg, f.inPath, f.outPath)
 		if err != nil {
-			return err
+			errs = multierror.Append(errs, err)
+			continue
 		}
 		if diff != "" {
 			if verbose {
@@ -76,6 +79,9 @@ func runFmt(inDir, outDir string, verbose bool) error {
 	if changes != "" {
 	}
+	if err := errs.ErrorOrNil(); err != nil {
+		return err
+	}
 	return nil
 }

```

nywilken (changes requested):
@kytrinyx this is just about ready to go. I was doing a full round of testing on the changes and found a bug in the runFmt function. The function returns on the first error, but it does not display any file changes that may have happened before the error occurred. I left the reproduction steps, a recommendation, and a quick patch (feel free to change as you see fit) in my comments.

kytrinyx:
> I was doing a full round of testing on the changes and found a bug in the runFmt function.

Thank you so much for catching this!

kytrinyx:
@nywilken I made those changes and tested them with the reproduction steps that you had 🌸

track/config.go
```diff
@@ -44,3 +74,34 @@ func NewConfig(path string) (Config, error) {
 	}
 	return c, nil
 }
+
+// NewConfigFromFile loads a config from file given the path to the file.
+func (cfg *Config) NewConfigFromFile(path string) error {
```
nywilken: Reading it over one more time, maybe the name should be `LoadFromFile` ☮️ ☮️

nywilken (approved):
@kytrinyx 💥 this is good to ship. I left a comment on the naming of the NewConfigFromFile method, but now I’m just nitpicking lol. I’m currently mobile reviewing so could you squash and merge, please. Otherwise I can handle it later when behind a machine.

track/config.go
kytrinyx: Heh. I'm ok with either. Let me think about this for an hour or so to see if I lean more strongly in either direction :)

kytrinyx: Ok, I decided I like LoadFromFile better :)

kytrinyx:
I will merge and cut a release tonight 🎉 Thanks for your help!

track/config.go
nywilken: 😃 👍

PR #138, opened 2018-06-11
Make checklist issue an integer

### LINKED ISSUE #136: configlet lint raises error if `checklist_issue` contains a number
Saw the issue flagged in https://github.com/exercism/powershell/issues/69 and was having the same problem on perl6. If `checklist_issue` contains a number the following error occurs:
`-> invalid config config.json -- json: cannot unmarshal number into Go struct field Config.checklist_issue of type string`.

This is new to `v3.9.1` and does not occur in `v3.9.0`.

kytrinyx:
The value is numeric; a string just doesn't make much sense here.

I've not added a test for this, since I can't see what kind of problem such a test would catch (if I had added a test for this previously, I would have assumed it was a string).

Closes #136 

We will need to fix the config in http://github.com/exercism/powershell when we cut the release for this.

### DIFF
--- a/track/config.go
+++ b/track/config.go
@@ -43,7 +43,7 @@ type Config struct {
 	Active         bool   `json:"active"`
 	Blurb          string `json:"blurb"`
 	Gitter         string `json:"gitter,omitempty"`
-	ChecklistIssue string `json:"checklist_issue,omitempty"`
+	ChecklistIssue int    `json:"checklist_issue,omitempty"`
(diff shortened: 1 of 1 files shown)


nywilken (approved):
All good here.

kytrinyx:
Faster than a speeding bullet 😂 Thanks so much! I'm hoping to cut a release for the fix today so we can wrap all of this up.

PR #143, opened 2018-07-15
WIP Add Blurb check to lint command

nywilken:
This change introduces a new check to the lint command that will validate that a Track's config has a blurb with at least 10 words. This check does not verify the quality of the said blurb.

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -153,7 +153,7 @@ func lintTrack(path string) bool {
 
 func incompleteBlurb(t track.Track) []string {
 	var tracks []string
-	re := regexp.MustCompile(`(?im:\b[\w]{5,}\b)`)
+	re := regexp.MustCompile(`(?im:\b[\w]{4,}\b)`)
 	words := re.FindAllString(t.Config.Blurb, -1)
 	if len(words) < 10 {
 		tracks = append(tracks, t.Config.TrackID)
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/exercism/configlet/track"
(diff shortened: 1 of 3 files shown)


cmd/lint_test.go
```diff
@@ -22,52 +19,86 @@ var (
...
-	defer func() {
-		ui.Out = originalOut
-		ui.ErrOut = originalErrOut
-	}()
-
-	lintTests := []struct {
-		desc     string
-		path     string
-		expected bool
+//func TestLintTrack(t *testing.T) {
```
kytrinyx, line 22: Should we be deleting rather than commenting out the test?

cmd/lint.go
```diff
@@ -146,6 +151,16 @@ func lintTrack(path string) bool {
 	return hasErrors
 }
 
+func incompleteBlurb(t track.Track) []string {
+	var tracks []string
+	re := regexp.MustCompile(`(?im:\b[\w]{4,}\b)`)
+	words := re.FindAllString(t.Config.Blurb, -1)
+	if len(words) < 10 {
```
kytrinyx, line 158: The error message says 30, this says 10 (and the PR says 8)... should we make a constant that both places can refer to?

cmd/lint_test.go
```diff
@@ -363,9 +394,9 @@ func TestLockedCoreViolation(t *testing.T) {
 			assert.Equal(t, tt.expectedSlug, slugs[0], tt.desc)
 		}
 	}
-}
+} // }}}
```
kytrinyx, line 397: What are these comments for?

fixtures/numbers/config.json
```diff
@@ -3,6 +3,7 @@
   "language": "Numbers",
   "repository": "https://github.com/exercism/numbers",
   "active": true,
+   "blurb": "Ruby is a dynamic, open source programming language with a focus on simplicity and productivity. It has an elegant syntax that is natural to read and easy to write.",
```
kytrinyx, line 6: Let's make this stupider... like "One two three four five..." up to however many words we have required.

kytrinyx (commented):
Overall this looks great. I have a couple of small comments/questions.

cmd/lint_test.go
nywilken: Oh forgot to to uncomment after my testing 🤦‍♂️

cmd/lint_test.go
nywilken: Oh sorry my vim markers for code folding 🤪

fixtures/numbers/config.json
nywilken: Sounds good. Thanks for taking a look over this PR.

nywilken:
Given its age I think it’s safe to close and revisit if there is still a need.

Issue #148, opened 2018-08-10
Add a --diff flag for `fmt` that will not actually rewrite the files


kytrinyx:
This should return a non-zero code if there is a diff.

See https://github.com/exercism/meta/issues/105

ekingery:
Closed by PR #150

PR #150, opened 2018-10-20
implement the --diff flag for the fmt command

danielj-jordan:
#148 

I am a bit new to Go and eager for any comments.  Thanks

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -13,8 +13,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// verbose flag for fmt command.
-var fmtVerbose bool
+var (
+	// verbose flag for fmt command.
+	fmtVerbose bool
+
+	// test flag for fmt command displays the proposed changes
+	fmtTest bool
+)
 
 // fmtCmd defines the fmt command
 var fmtCmd = &cobra.Command{
@@ -29,18 +34,26 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
 `,
 	Example: fmt.Sprintf("  %s fmt %s --verbose", binaryName, pathExample),
 	Run: func(cmd *cobra.Command, args []string) {
-		if err := runFmt(args[0], args[0], fmtVerbose); err != nil {
+
+		if diffFound, err := runFmt(args[0], args[0]); err != nil {
 			ui.PrintError(err.Error())
 			os.Exit(1)
+		} else if diffFound && fmtTest {
+			os.Exit(2)
 		}
+
 	},
 
 	Args: cobra.ExactArgs(1),
 }
 
-func runFmt(inDir, outDir string, verbose bool) error {
+func runFmt(inDir, outDir string) (bool, error) {
 	if _, err := os.Stat(filepath.Join(outDir, "config")); os.IsNotExist(err) {
 		os.Mkdir(filepath.Join(outDir, "config"), os.ModePerm)
+	}
+
+	if fmtTest {
+		fmtVerbose = true
 	}
 
 	var fs = []struct {
@@ -61,6 +74,7 @@ func runFmt(inDir, outDir string, verbose bool) error {
 	}
 
 	var changes string
+	var diffFound bool
 
 	errs := &multierror.Error{}
 	for _, f := range fs {
@@ -70,19 +84,25 @@ func runFmt(inDir, outDir string, verbose bool) error {
 			continue
 		}
 		if diff != "" {
-			if verbose {
+			if fmtVerbose {
 				ui.Print(fmt.Sprintf("%s\n\n%s", f.inPath, diff))
 			}
 			changes += fmt.Sprintf("%s\n", f.inPath)
+			diffFound = true
+
 		}
 	}
 	if changes != "" {
-		ui.Print("changes made to:\n", changes)
+		if fmtTest {
+			ui.Print("no changes were made to:\n", changes)
+		} else {
+			ui.Print("changes made to:\n", changes)
+		}
 	}
 	if err := errs.ErrorOrNil(); err != nil {
-		return err
+		return diffFound, err
 	}
-	return nil
+	return diffFound, nil
 }
 
 func formatFile(cfg ConfigSerializer, inPath, outPath string) (string, error) {
@@ -105,7 +125,7 @@ func formatFile(cfg ConfigSerializer, inPath, outPath string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if diff != "" {
+	if diff != "" && !fmtTest {
 		if err := ioutil.WriteFile(outPath, dst, os.FileMode(0644)); err != nil {
 			return "", err
 		}
@@ -116,4 +136,5 @@ func formatFile(cfg ConfigSerializer, inPath, outPath string) (string, error) {
 func init() {
 	RootCmd.AddCommand(fmtCmd)
 	fmtCmd.Flags().BoolVarP(&fmtVerbose, "verbose", "v", false, "display the diff of the formatted changes.")
+	fmtCmd.Flags().BoolVarP(&fmtTest, "test", "t", false, "display the proposed changes, but do not make them.")
 }
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -74,7 +74,6 @@ func runFmt(inDir, outDir string) (bool, error) {
 	}
 
 	var changes string
-	var diffFound bool
 
 	errs := &multierror.Error{}
 	for _, f := range fs {
@@ -88,11 +87,11 @@ func runFmt(inDir, outDir string) (bool, error) {
 				ui.Print(fmt.Sprintf("%s\n\n%s", f.inPath, diff))
 			}
 			changes += fmt.Sprintf("%s\n", f.inPath)
-			diffFound = true
 
 		}
 	}
-	if changes != "" {
+	diffFound := changes != ""
+	if diffFound {
 		if fmtTest {
 			ui.Print("no changes were made to:\n", changes)
 		} else {
--- a/cmd/fmt_example_test.go
+++ b/cmd/fmt_example_test.go
@@ -8,7 +8,7 @@ import (
 	"github.com/exercism/configlet/ui"
 )
 
-func ExampleFormat() {
+func ExampleFormatVerboseFlag() {
 	oldOut := ui.Out
 	oldErrOut := ui.ErrOut
 	ui.Out = os.Stdout
@@ -24,7 +24,10 @@ func ExampleFormat() {
 	}
 	defer os.Remove(unformattedDir)
 
-	runFmt("../fixtures/format/unformatted/", unformattedDir, true)
+	fmtVerbose = true
+	fmtTest = false
+
+	runFmt("../fixtures/format/unformatted/", unformattedDir)
 
 	// Output:
 	//-> ../fixtures/format/unformatted/config.json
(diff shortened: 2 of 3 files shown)


cmd/fmt.go
```diff
@@ -29,19 +32,24 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
...
 
 	Args: cobra.ExactArgs(1),
 }
 
-func runFmt(inDir, outDir string, verbose bool) error {
+func runFmt(inDir, outDir string, verbose bool, diffOnly bool) (bool, error) {
 	if _, err := os.Stat(filepath.Join(outDir, "config")); os.IsNotExist(err) {
 		os.Mkdir(filepath.Join(outDir, "config"), os.ModePerm)
 	}
+	var diffFound = false
```
nywilken: When assigning the zero value (default value) for a type it is preferred to just use `var diffFound bool`

cmd/fmt.go
```diff
@@ -29,19 +32,24 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
 `,
 	Example: fmt.Sprintf("  %s fmt %s --verbose", binaryName, pathExample),
 	Run: func(cmd *cobra.Command, args []string) {
-		if err := runFmt(args[0], args[0], fmtVerbose); err != nil {
+
+		if diffFound, err := runFmt(args[0], args[0], fmtVerbose, fmtDiff); err != nil {
```
nywilken: Seeing as `fmtVerbose` and `fmtDiff` are package level variables we should be able to access them from within `runFmt` without having to pass them in.

If you remove the parameters from here and within the `runFmt` function you would need to update the references  of `verbose` and `diffonly` to `fmtVerbose` and `fmtDiff`

cmd/fmt_example_test.go
```diff
@@ -24,7 +24,7 @@ func ExampleFormat() {
 	}
 	defer os.Remove(unformattedDir)
 
-	runFmt("../fixtures/format/unformatted/", unformattedDir, true)
+	runFmt("../fixtures/format/unformatted/", unformattedDir, true, false)
```
nywilken: If you make the update so that `runFmt` uses `fmtVerbose` and `fmtDiff` directly you could remove the parameters and instead use the convention on lines 12-19 to set and reset the values for the pkg level variables `fmtVerbose` and `fmtDiff`. 

```
oldVerbose = fmtVerbose
fmtVerbose = true
defer func() {
  fmtVerbose = oldVerbose
}
```

cmd/fmt_test.go
```diff
@@ -26,7 +26,7 @@ func TestFmtCommand(t *testing.T) {
 		t.Fatal(err)
 	}
 	defer os.Remove(formattedDir)
-	runFmt("../fixtures/format/formatted/", formattedDir, false)
+	runFmt("../fixtures/format/formatted/", formattedDir, false, false)
```
nywilken: Same comment I left for fmt_example_test applies here.

nywilken (dismissed):
@danielj-jordan I took a quick pass-through and left a few comments pertaining to the `runFmt` function signature. I will continue looking through the PR and make suggestions where possible.

Thanks again for opening up the PR!

cmd/fmt_test.go
danielj-jordan: Thanks @nywilken.  I made the changes that you suggested.   If you like the current code, I can squash and re-submit the PR if you like.

cmd/fmt.go
```diff
@@ -70,19 +81,25 @@ func runFmt(inDir, outDir string, verbose bool) error {
...
 				ui.Print(fmt.Sprintf("%s\n\n%s", f.inPath, diff))
 			}
 			changes += fmt.Sprintf("%s\n", f.inPath)
+			diffFound = true
+
 		}
 	}
 	if changes != "" {
-		ui.Print("changes made to:\n", changes)
+		if fmtDiff == false {
```
kytrinyx: Here we basically have an "if not else".

I think it would make sense to swap it around so that it reads `if fmtDiff, otherwise..."

cmd/fmt.go
```diff
@@ -105,7 +122,7 @@ func formatFile(cfg ConfigSerializer, inPath, outPath string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if diff != "" {
+	if diff != "" && fmtDiff == false {
```
kytrinyx: It would be more idiomatic to make this `!fmtDiff`

cmd/fmt.go
```diff
@@ -116,4 +133,5 @@ func formatFile(cfg ConfigSerializer, inPath, outPath string) (string, error) {
 func init() {
 	RootCmd.AddCommand(fmtCmd)
 	fmtCmd.Flags().BoolVarP(&fmtVerbose, "verbose", "v", false, "display the diff of the formatted changes.")
+	fmtCmd.Flags().BoolVarP(&fmtDiff, "diff", "d", false, "display the diff of the proposed changes, but do not make them.")
```
kytrinyx: How would you feel about `dry-run` as the flag name instead of `diff`?
That's a very common flag name for something that would try something out, report what would happen, but then not actually make the changes.

@nywilken I'd love to hear your thoughts on this one as well.

cmd/fmt_example_test.go
```diff
@@ -24,7 +24,10 @@ func ExampleFormat() {
 	}
 	defer os.Remove(unformattedDir)
 
-	runFmt("../fixtures/format/unformatted/", unformattedDir, true)
+	fmtVerbose = true
+	fmtDiff =  false
+
+	runFmt("../fixtures/format/unformatted/", unformattedDir)
```
kytrinyx, line 30: I really like how much more readable this became now that the booleans are being assigned to the package variables!

kytrinyx (changes requested):
This is looking great! I have a couple of tiny comments/suggestions, but I think this is basically good to go.

cmd/fmt.go
danielj-jordan: I think it is better named dry-run also.

danielj-jordan:
@nywilken @kytrinyx Thanks for the feedback.  I have incorporated all your recommendations and squashed the commit.

cmd/fmt.go
nywilken: > How would you feel about `dry-run` as the flag name instead of diff?

:+1: I like that option over `diff`. I would even go so far as saying that we copy the option and messaging from the grep command 
 ```
-n, --dry-run       perform a trial run with no changes made
```

Comment (nywilken) [reply cmd/fmt.go]: Actually, looking at the original issue we want a non-zero code if there are formatting issues so a `dry-run` might not be the option we want. In fact, I wouldn't expect `dry-run` to exit without error if the command successfully prints out the changes that would be written to the source file. 

What if we keep the `dry-run` option, but also add a `-t, --test` flag that returns a non-zero if there are formatting issue? 

`-t, —test performs a trial and exits if formatting is needed; meant to be used with ci`

Comment (nywilken) [reply cmd/fmt.go]: >  What if we keep the `dry-run` option, but also add a `-t, --test` flag that returns a non-zero if there are formatting issue. 

I pulled down the branch and see that dry-run does not actually print the changes unless the verbose flag is passed as well so I would like to change my previous suggestion slightly. Which of course is open for discussion. 

Rename the `dry-run` option to `test` or `check` using the messaging above, enable verbosity any time the check flag is passed so that if the check fails users will see the diff. Thoughts?

File: cmd/fmt.go [hunk 9]
```diff
@@ -70,19 +81,25 @@ func runFmt(inDir, outDir string, verbose bool) error {
...
-			if verbose {
+			if fmtVerbose {
 				ui.Print(fmt.Sprintf("%s\n\n%s", f.inPath, diff))
 			}
 			changes += fmt.Sprintf("%s\n", f.inPath)
+			diffFound = true
+
 		}
 	}
 	if changes != "" {
```
nywilken: The declaration of `diffFound` can actually be moved right above this line with a slight change

```suggestion
diffFound := changes != ""
if diffFound {
...
```

cmd/fmt.go
```diff
@@ -29,19 +35,24 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
...
 
 	Args: cobra.ExactArgs(1),
 }
 
-func runFmt(inDir, outDir string, verbose bool) error {
+func runFmt(inDir, outDir string )(bool, error) {
 	if _, err := os.Stat(filepath.Join(outDir, "config")); os.IsNotExist(err) {
 		os.Mkdir(filepath.Join(outDir, "config"), os.ModePerm)
 	}
+	var diffFound bool
```
nywilken: This declaration can be made closer to where it is being used.

cmd/fmt.go
```diff
@@ -70,19 +81,25 @@ func runFmt(inDir, outDir string, verbose bool) error {
 			continue
 		}
 		if diff != "" {
-			if verbose {
+			if fmtVerbose {
 				ui.Print(fmt.Sprintf("%s\n\n%s", f.inPath, diff))
 			}
 			changes += fmt.Sprintf("%s\n", f.inPath)
+			diffFound = true
```
nywilken: If you apply the suggestions made below you can drop this line.

cmd/fmt.go
```diff
@@ -70,19 +81,25 @@ func runFmt(inDir, outDir string, verbose bool) error {
...
+
 		}
 	}
 	if changes != "" {
-		ui.Print("changes made to:\n", changes)
+		if fmtDryRun {
```
nywilken: Seeing as the flag description specifies that no changes will be made I would drop this if/else and instead only print when changes were made. 

```
if diffFound && !fmtDryRun {
}
```

cmd/fmt.go
```diff
@@ -70,19 +81,25 @@ func runFmt(inDir, outDir string, verbose bool) error {
...
+			ui.Print("no changes were made to:\n", changes)
+		} else {
+			ui.Print("changes made to:\n", changes)
+		}
 	}
 	if err := errs.ErrorOrNil(); err != nil {
-		return err
+		return diffFound, err
```
nywilken, line 102: If you apply the suggestion on line 92 you can just return `false` with the error

cmd/fmt.go
```diff
@@ -29,19 +35,24 @@ It also normalizes and alphabetizes the exercise topics in the config.json file.
 `,
 	Example: fmt.Sprintf("  %s fmt %s --verbose", binaryName, pathExample),
 	Run: func(cmd *cobra.Command, args []string) {
-		if err := runFmt(args[0], args[0], fmtVerbose); err != nil {
+
```
nywilken, line 37: What do you think enabling verbosity anytime the format check is run? Something like the following right before we call `runFmt(...)` 

```
if fmtDryRun {
    fmtVerbose = true
}

```

nywilken (commented):
@danielj-jordan @kytrinyx this is coming along nicely. I left a couple of suggestions along with my thoughts on the flag name. FWIW the code suggestions within this review should not have to change if we decide to rename the flag - only the references to the flag variable.

cmd/fmt.go
kytrinyx: @nywilken these are good observations. I think that it's important that we can exit with an error code in order to be able to use this in CI. In that sense, I think that `test` communicates this better than `dry-run`.

For the doctor command, I think we _always_ need to print in the test version, whether or not `verbose` is passed. I think it's fine to not print verbose output if just running doctor to fix something.

cmd/fmt.go
danielj-jordan: Yes.  It makes sense to do this

danielj-jordan:
I want to confirm that the flag should be named —test not —dry-run.   Please reply and confirm.   Thank you

kytrinyx:
@danielj-jordan thank you for confirming. Yes! We're going with `--test` rather than `--dry-run`.

Also, the command should

1. set verbose to true even if verbose wasn't passed, and
2. exit with a non-zero code if something needs changing

danielj-jordan:
@kytrinyx @nywilken I am not sure if you got notified, but the —test flag is implemented.   It always runs with verbose.

kytrinyx:
@danielj-jordan Ah, nope! I didn't see the notification. Thanks so much, I'll take a look at this in the morning.

cmd/fmt.go
kytrinyx: This should still be moved closer to where it's used 👍

kytrinyx (changes requested):
There's one more minor tweak to do (the `diffFound` thing that @nywilken mentioned previously. After that I think this is good to go!

danielj-jordan:
@kytrinyx  I moved the declaration closer.  Thanks

kytrinyx (approved):
This is great, thanks so much @danielj-jordan!

ekingery:
Closes #148

PR #151, opened 2018-11-26
scripts: add fetch-configlet scripts for Unix and Windows

ErikSchierboom:
Many repositories have a `fetch-configlet` script in their `bin` folder, but as far as I can tell no such file is maintained in this repository. To me, this goes against the principle of least surprise, so this PR adds that script in this repository. I've also created a version of the fetch-configlet script that runs on Windows (which uses PowerShell). The original `fetch-configlet` script made it seem like it also supported Windows, but it did not work on Windows.

I'm not entirely sure this is the right place for this PR. Let me know if it's not.

### DIFF
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+LATEST=https://github.com/exercism/configlet/releases/latest
+
+OS=$(
+case $(uname) in
+    (Darwin*)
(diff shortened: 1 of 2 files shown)


kytrinyx (approved):
Brilliant! Yes this is absolutely the right place for this script. Thank you!

ErikSchierboom:
🎉

PR #158, opened 2019-04-29
Migrated the project to the Go modules for dependency management

ZapAnton:
Migrated the project to use the Go modules for dependency management, similar to the https://github.com/exercism/cli/pull/812.


ZapAnton:
For some reason Travis did not fail, even though the checked versions of Go are `1.8` and `1.9`, which do not have the go modules support. Am I missing something?

PR #161, opened 2019-05-10
fetch-configlet.sh: Fetch the latest release url from the Github API

ZapAnton:
This PR changes the approach the URL for the latest release of `configlet` is built.
Before the PR: Manually build the URL using the version from the HTTP header.
After the PR: Get the full URL from the [Github API](https://developer.github.com/v3/repos/releases/#get-the-latest-release). IMHO this approach is more reliable.

### DIFF
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -37,8 +37,7 @@ case $(uname -m) in
 esac)
 
 
-VERSION="$(curl --silent --head $LATEST | awk -v FS=/ '/Location:/{print $NF}' | tr -d '\r')"
-URL=https://github.com/exercism/configlet/releases/download/$VERSION/configlet-$OS-${ARCH}.$EXT
+URL=$(curl -s https://api.github.com/repos/exercism/configlet/releases/latest | awk "/browser_download/ && /$OS-$ARCH.$EXT/" | cut -d : -f 2,3 | tr -d \")
 
 case $EXT in
     (*zip)


SleeplessByte:
How exactly is it more reliable?

ZapAnton:
By moving from manually building the URL, which may be error-prone if the script is further expanded, to making the Github build the URL for us making it the single source of truth.
Plus parsing the header of the HTTP response header seems a bit awkward - can you guarantee that the header structure will not change between the calls? The same could be asked about the Github API, but I have an impression that the returned JSON is still more stable than a header.

SleeplessByte:
> Can you guarantee that the header structure will not change between the calls?

Since GitHub is using REST urls, I think it's equally unlikely for the URL to `latest` to change as the API to change. The header is the `Location` header, which structure has been the same, [at least since 1999](https://tools.ietf.org/html/rfc2616?spm=5176.doc32013.2.3.Aimyd7#section-14.30), a header which your browser actually relies on to redirect you to the next page (which is what happens if you browse to https://github.com/exercism/configlet/releases/latest).

If you're asking if we can guarantee if the _value_ of that header will always be actually the latest version (i.e. redirect to tag/latest-version); the answer to that is "about as well as a guarantee the API will not change". 

> The same could be asked about the Github API, but I have an impression that the returned JSON is still more stable than a header.

I think there is no proof to support that impression.

Note that in _both_ cases we're completely relying on the author of the tag to:
- name the tag correctly
- name the files correctly

and in both cases we're looking for the correct url; right now by building it, or after this PR by searching for it. Both will fail if the tag author has failed.

I personally don't feel this is an improvement to the script (but also not a declination). The pro for your change would be that you're only making _1 request_ instead of 2; which ís somewhat favourable.

PR #163, opened 2019-07-07
Allow a DO_NOT_IMPLEMENT.md file in foregone exercises

jmrunkle:
Modify the configlet code to allow use-cases like https://github.com/exercism/java/pull/1711. Basically, by allowing this one file in the stub directory, we can make it clearer if there are reasons (for example, a license problem) that we cannot implement the exercise. The markdown file would also allow us to explain why the exercise cannot be implemented in this track.

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -270,7 +270,7 @@ func foregoneViolations(t track.Track) []string {
 
 	slugs := []string{}
 	for _, exercise := range t.Exercises {
-		if violations[exercise.Slug] {
+		if violations[exercise.Slug] && !exercise.IsDoNotImplementOnly() {
 			slugs = append(slugs, exercise.Slug)
 		}
 	}
--- a/cmd/lint_test.go
+++ b/cmd/lint_test.go
@@ -224,7 +224,7 @@ func TestForegoneViolationWithDoNotImplementOnly(t *testing.T) {
 	slugs := foregoneViolations(track)
 
 	if len(slugs) != 0 {
-		t.Fatalf("Expected no foregone violations in 0 exercises, but found violations: %s", string(slugs))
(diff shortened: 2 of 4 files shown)


jmrunkle:
@kytrinyx - is there someone I should contact about this pull request?

kytrinyx:
Hi @jmrunkle, thanks so much for submitting this. I am traveling for work until Friday, and I will need to catch up on this on the weekend, unless I can sneak away for a bit before that.

kytrinyx:
@jmrunkle Before I review this though, would you mind pointing to the issue where this problem was discussed, and the various approaches to solving it were evaluated? I'd like to see what went into the design decisions for this before I look at implementation questions. (I couldn't find a discussion, I would have expected to see it in http://github.com/exercism/exercism/issues somewhere).

jmrunkle:
See some discussion here: https://github.com/exercism/java/issues/1589

Basically, the foregone issues list in the config gives a late signal to the developer that they should not implement the exercise (only once they have already done the work and created a PR).

With this solution, there would be a file in the directory which stands as a clear indicator that they should not implement the exercises right when they go to start implementing the exercise (and it would contain the reason why).

jmrunkle:
Just opened https://github.com/exercism/exercism/issues/4932 for discussion by all track maintainers (and commented on the slack channel so that hopefully they will see it there as well).

yawpitch:
The motivation for this ended with #1712 being merged in late July. See my comment [here](https://github.com/exercism/exercism/issues/4932#issuecomment-552202038).

yawpitch:
Since exercism/exercism#4932 has been closed, this probably should be as well.

kytrinyx:
Thanks to everyone who participated in finding a way forward on this one!

PR #166, opened 2019-08-06
fetch-configlet: set pipefail shell option

guygastineau:
So here it is with the `pipefail` option set.

I usually stay away from the `errexit` option, but it occurs to me that it could be beneficial in this case.
If we also set the the `errexit` option then I believe this script will exit with an appropriate error message even if the subprocess from the command substitution in the assignment to `VERSION` fails.  This does seem like the best behavior.

Therefore, please let me know if there is agreement that `errexit` should also be set :smile:

PS. I have a style concern about the terminators in the case statements.  This is outside of this PR's scope, but after this is merged I will open a PR that presents changes to the case style (specifically to make it conform more closely to some of the most accepted Bash style guides on the web ;) )

### DIFF
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -o pipefail
+
 LATEST=https://github.com/exercism/configlet/releases/latest
 
 OS=$(


kytrinyx:
> Therefore, please let me know if there is agreement that `errexit` should also be set

LOL. You're assuming so much about my knowledge of exit codes. I'm in the "seems fine" group :-)

kytrinyx (approved):
Thank you so much for bailing me out of my very limited bash knowledge ❤️

kytrinyx:
> I have a style concern

I would gladly accept style concern fixes.

Once you've updated that, I'll update the various track builds to get this file from this repo, instead of storing it, I think.

guygastineau:
Cool. I didn't set the `errexit` option yet, but that is trivial. It will literally requires the addition of 1 character, so I will just include it in the style PR if that is acceptable 😀

I am taking my mom to the shoe store, so it'll be later today before I get the next PR up here 😎

kytrinyx:
> I will just include it in the style PR if that is acceptable

Totally! And I see that shoe mission is complete, and style PR is up :-) I'll go look at that now.

Issue #99, opened 2017-10-15
configlet lint: Should check whether UUIDs are valid.


petertseng:
**Decision**: `configlet lint` should check whether UUIDs are valid.

**Support**:

* https://github.com/exercism/configlet/issues/99#issuecomment-349434007
* https://github.com/exercism/configlet/issues/99#issuecomment-350134869

**Opposition**: None.

**Task**: Make `configlet lint` check whether UUIDs are valid.

---

**Original issue text/question**: 

What does valid mean? Is it the standard `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` where each x is a hexadecimal digit?

Whatever "valid" means in the Exercism context, would it be good for configlet to check for invalid UUIDs? If whatever is using those UUIDs is expecting them all to be valid, it might be needed.

```
[gems/trackler-2.2.1.45]$ for c in tracks/*/config.json; do echo $c; ruby -rjson -e 'h = /[0-9a-fA-F]/; j = JSON.parse(ARGF.read); j["exercises"].each { |e| u = e["uuid"]; puts "%24s has %s" % [e["slug"], u] if u !~ /^#{h}{8}-#{h}{4}-#{h}{4}-#{h}{4}-#{h}{12}$/ }' $c; done 
tracks/bash/config.json
tracks/c/config.json
tracks/ceylon/config.json
tracks/cfml/config.json
tracks/clojure/config.json
tracks/coffeescript/config.json
tracks/common-lisp/config.json
tracks/coq/config.json
tracks/cpp/config.json
           atbash-cipher has ba4e0186-0389-af80-d3e4-695dfc4e63bff1fa503
            bracket-push has ce9fd8ac-0beb-1a80-4ff6-be19ce1540d905c31b7
                 pangram has ea8d6224-0440-6d80-1569-a04127908a200d2ecf0
tracks/crystal/config.json
tracks/csharp/config.json
tracks/dart/config.json
               raindrops has d6657d4a-0257-0c80-6667-fddd127a571f0ad1fd4
tracks/d/config.json
tracks/delphi/config.json
             hello-world has 765b5bcf-00e0-d180-ea7f-92dae26e6da0b30151c
                 two-fer has b38f504c-098e-fe80-896f-60e6c45751033d26161
                     bob has cd80ef91-0451-c380-3d58-01a2ddff3029f430eea
                    leap has 9234d900-050c-5980-561b-e1f772ae2b9c7b3fd08
[...]

PR #168, opened 2019-08-26
lint: Check that UUIDs are valid

petertseng:
all tests that involve lint are hereby updated to do one of the two:
1. have valid UUIDs
2. keep their invalid UUIDs, and expect that an error for them is shown.

Closes https://github.com/exercism/configlet/issues/99

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/google/uuid"
 	"github.com/exercism/configlet/track"
 	"github.com/exercism/configlet/ui"
 	"github.com/spf13/cobra"
@@ -104,6 +105,10 @@ func lintTrack(path string) bool {
 		{
 			check: missingUUID,
 			msg:   "The exercise '%v' was found in config.json, but does not have a UUID.",
+		},
+		{
+			check: invalidUUID,
+			msg:   "The exercise '%v' has an invalid UUID in config.json.",
 		},
 		{
 			check: foregoneViolations,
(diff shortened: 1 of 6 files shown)


petertseng:
Note this list of exercises and their associated tracks that currently have invalid UUIDs:

invalid elm isogram: 5f540090-061e-2f80-40a8-d9782700ed2efdf8965: invalid UUID length: 43
invalid elisp pangram: 28aea3bc-0ca0-4af9-868a-02e71c3d996: invalid UUID length: 35
invalid cpp all-your-base: 0a3220c6-07a9-ed80-4d2d-1fc65de969b771a46d5: invalid UUID length: 43
invalid bash queen-attack: text_formatting: invalid UUID length: 15
invalid python custom-set: bb07c236-062c-2980-483a-a221e4724445dcd6f32: invalid UUID length: 43
invalid plsql hello-world: 53156fc9-0945-f080-08a9-ce269bfc9121a5d2c5c: invalid UUID length: 43
invalid clojure spiral-matrix: d42g42d7-21b1-4daf-8ddc-a2ab9ee07c98: invalid UUID format
invalid r diamond: 7cad82d6-0db6-d680-4133-81d850d6e7a925454a0: invalid UUID length: 43
invalid haxe leap: d30952cf-0b7d-a980-9351-caedebfed16362c6a7a: invalid UUID length: 43

kytrinyx:
We need to figure out what would have to happen in the database to change UUIDs to be valid, where they are not. Currently, fixing invalid UUIDs breaks the site pretty badly.

petertseng:
It looks like this PR will have to wait until those issues are solved.

I sympathise with this plight - it means that it would currently be unwise to release a configlet with this change - affected tracks' `configlet lint` would break, and you've indicated that fixing their UUIDs is not the way to go. It would be frustrating to be in that situation with no way out.

The motivation for having this PR is to prevent any additional invalid UUIDs from being added. Since we see that configlet currently can't be the place where this is enforced, I will do so in individual tracks' CI of tracks I maintain. Thus achieving the original motivation without causing trouble for all tracks and the website.

When the above issues are solved and configlet is able to enforce this, then it will be a good time to revisit this PR and remove the checks from individual tracks.

Issue #140, opened 2018-06-12
configlet fmt on deprecated exercises: Should not re-add `core`, `unlocked_by`, `difficulty`, `topics`


petertseng:
recent `configlet fmt` adds these keys to all exercises in the `deprecated` array

```diff
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
```

In https://github.com/exercism/discussions/issues/159#issuecomment-315466131 I argued that I saw no reason to keep `difficulty` and `topics` there. I continue to see no reason, however this is your chance to make me see a reason.

For example, perhaps any exercise that becomes deprecated could simply retain all its old values for these fields. And perhaps that is eminently useful, for some reason. That would certainly be cause for me to see a reason.

In such a case, please be advised that I'm not going to expend mental energy on making up values for exercises that were deprecated before this operation, so these values are going to remain whatever values `configlet fmt` decided to place for them (`0` and `null` respectively), and I hope that's okay.

I also reject a workflow that manually removes these keys from deprecated exercises after having run `configlet fmt` because 1) why would any person want to do manual work 2) https://github.com/exercism/haskell/blob/master/bin/check-configlet-fmt.sh https://github.com/exercism/rust/blob/master/_test/check-configlet-fmt.sh would make Travis CI reject it anyway.

It may be possible to exclude those fields from the deprecated array only, but that would involve either keeping two structs in sync painstakingly, or some embedding, and I have not yet speculated on the clarity of the resulting structs. I cannot advise on this course of action.

If something else should happen, please advise.

kytrinyx:
> I continue to see no reason, however this is your chance to make me see a reason.

I agree with you. I had forgotten to add tests for this case.

PR #169, opened 2019-09-17
config: Only marshal Slug, UUID, Deprecated of deprecated exercises

petertseng:
We don't need any of the other fields, and it's deceptive to keep them
in the JSON file because a maintainer may mistakenly believe them to
be significant, but in fact they are not; the website does not display
any of the removed fields:
https://github.com/exercism/website/blob/b3314e84c52b52cf5f253386c3f0bf4127ca09aa/app/views/my/tracks/_show_side_exercises.html.haml#L26-L34

Two new tests were added: TestMarshalActive and TestMarshalDeprecated.
TestMarshalActive passes both before and after the associated code
change.
TestMarshalDeprecated fails before the associated code change, and
passes after.

Closes https://github.com/exercism/configlet/issues/140

### DIFF
--- a/track/config.go
+++ b/track/config.go
@@ -100,6 +100,36 @@ func (cfg *Config) ToJSON() ([]byte, error) {
 	return json.MarshalIndent(&cfg, "", "  ")
 }
 
+// MarshalJSON marshals a exercise metadata to JSON,
+// only marshalling certain fields if the exercise is deprecated.
+func (e *ExerciseMetadata) MarshalJSON() ([]byte, error) {
+	if e.IsDeprecated {
+		// Only marshal Slug, UUID, Deprecated.
+		return json.Marshal(&struct {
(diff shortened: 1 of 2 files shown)


petertseng:
todo: put https://github.com/exercism/website/blob/b3314e84c52b52cf5f253386c3f0bf4127ca09aa/app/views/my/tracks/_show_side_exercises.html.haml#L26-L34 in the commit message to show that on the website end none of the removed fields are shown

petertseng:
TODO complete, comment deleted.

petertseng:
I'd also consider not marshaling `unlocked_by` for `core: true` exercises but notice that this is a fundamentally different problem. This PR says "only marshal a specific set of fields". Not marshaling `unlocked_by` for `core` is "don't marshal a specific set of fields" which is harder to deal with. I'm not interested in making this change in this PR.

Issue #173, opened 2020-02-19
fetch-configlet - gzip: stdin: not in gzip format


amscotti:
I'm not really sure if this is an issue or just something I'm seeing due to something temporary or intermittent but when working on the Groovy track, when running `fetch-configlet` I'm getting this error,

```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

I have seen this locally, (Mac OS, Ubuntu on WSL) along with on CI (Travis and GitHub Actions). If I keep running the build it will eventually pass without any errors from `fetch-configlet`. 

[Link to failing build on Travis](https://travis-ci.org/exercism/groovy/builds/652182492)

Looking into this a bit, it seems that the script is unable to get the version and is using the wrong URL to download from. One of the things I have tried is upgrading `fetch-configlet` to the latest version but I was still seeing this issue.

I can no longer reproduce this locally, so maybe it's something just intermittent but I do still see this popping up in CI.

Let me know if I can do anything to help by providing more information, it is kind of tricky because it is intermittent but I'll do my best with adding any more details that I can. This does seem related to #165, but that seems to be more about reporting the error then fixing it. 

Thanks!

ErikSchierboom:
I have had the same issue for the C# and F# tracks.

ErikSchierboom:
This is a link to the failing build on GitHub actions: https://github.com/exercism/csharp/pull/1364/checks?check_run_id=465607250

nicuveo:
Same problem on the Haskell track; I have two PRs that won't build (https://github.com/exercism/haskell/pull/896 and https://github.com/exercism/haskell/pull/897).

smuroff:
The problem was [here](https://github.com/exercism/cpp/pull/344/files). CI [build](https://travis-ci.org/exercism/cpp/builds/657670576) passed on cpp track.

nicuveo:
This PR also proposes a fix that uses `perl` instead to make the check case insensitive: https://github.com/exercism/sml/pull/127.

JesseSingleton:
Having this issue when trying to PR into the Javascript track.

pclausen:
same issue on fortran track: https://github.com/exercism/fortran/issues/71

Cohen-Carlisle:
The most robust way to fix this is to find the line in your CI doing:
```
[...]

PR #174, opened 2020-03-12
Update scripts to support authentication

ErikSchierboom:
In issue #173, an issue was reported that reported that some runs of the `fetch-configlet` script led to the following error: "stdin: not in gzip format." As it turned out, this was due to the `fetch-configlet` script sometimes not being able to download the `configlet` binary, which is very likely due to rate limiting issues.

To work around this issue, this PR adds support for adding an authentication bearer token, which will increase the rate limit. It does this by checking for the presence of an environment variable named `GITHUB_TOKEN`. If such an environment variable is found, the authentication header is sent. Otherwise, a normal, unauthenticated request is made.

This issue will usually pop-up in CI scenarios, where the `configlet` script is fetched for each CI run. In the case when Github Actions is used as the CI, using this updated version of the `fetch-configlet` script is simple: simply update the contents of the `fetch-configlet` file to the one in this PR and add the following line to your GH Action workflow: https://github.com/exercism/csharp/pull/1375/files#diff-fe8421955fd596131bb6f1b78984b2fbR15 This will ensure that the automatically provided secret Github token value is passed along to the script.

If tracks use a different CI system, we'd have to create a token and then store that as a secret in the repositories. Let us know if you want to do this.

### DIFF
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -2,7 +2,7 @@
 
 set -eo pipefail
 
-readonly RELEASES='https://api.github.com/repos/exercism/configlet/releases/latest'
+readonly LATEST='https://api.github.com/repos/exercism/configlet/releases/latest'
 
 case "$(uname)" in
     (Darwin*)   OS='mac'     ;;
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -2,12 +2,7 @@
 
 set -eo pipefail
 
-readonly RELEASES='https://github.com/exercism/configlet/releases'
-
-get_version () {
-    curl --silent --head ${RELEASES}/latest  |
-        awk -v FS=/ '/Location:/{print $NF}' | tr -d '\r'
-}
+readonly RELEASES='https://api.github.com/repos/exercism/configlet/releases/latest'
 
 case "$(uname)" in
     (Darwin*)   OS='mac'     ;;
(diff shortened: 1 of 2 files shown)


ee7:
I'm glad to see that you solved this one!

However, I think the authentication changes to `fetch-configlet` should be based off the latest version of it that's in this repo.

The diff in this PR is currently large because it effectively reverts the changes from e.g. https://github.com/exercism/configlet/pull/167. It also removes Windows support.

ErikSchierboom:
@ee7 I've updated the PR. Let me know if it conforms to the correct style!

scripts/fetch-configlet
```diff
@@ -2,12 +2,7 @@
 
 set -eo pipefail
 
-readonly RELEASES='https://github.com/exercism/configlet/releases'
-
-get_version () {
-    curl --silent --head ${RELEASES}/latest  |
-        awk -v FS=/ '/Location:/{print $NF}' | tr -d '\r'
-}
+readonly RELEASES='https://api.github.com/repos/exercism/configlet/releases/latest'
```
ee7:
```suggestion
readonly LATEST='https://api.github.com/repos/exercism/configlet/releases/latest'
```
I guess we can change this back to `LATEST` again :)

scripts/fetch-configlet
```diff
@@ -30,15 +25,28 @@ case "$(uname -m)" in
...
+then
+    HEADER=''
+else
+    HEADER="authorization: Bearer ${GITHUB_TOKEN}"
+fi
+
+FILENAME="configlet-${OS}-${ARCH}.${EXT}"
+
+get_url () {
+    curl --header $HEADER -s $RELEASES |
```
ee7:
```suggestion
    curl --header "$HEADER" -s "$LATEST" |
```

scripts/fetch-configlet
```diff
@@ -30,15 +25,28 @@ case "$(uname -m)" in
...
+}
 
-VERSION="$(get_version)"
-URL="${RELEASES}/download/${VERSION}/configlet-${OS}-${ARCH}.${EXT}"
+URL=$(get_url)
 
 case "$EXT" in
     (*zip)
-        curl -s --location "$URL" -o bin/latest-configlet.zip
+        curl --header $HEADER -s --location "$URL" -o bin/latest-configlet.zip
```
ee7:
```suggestion
        curl --header "$HEADER" -s --location "$URL" -o bin/latest-configlet.zip
```

scripts/fetch-configlet
```diff
@@ -30,15 +25,28 @@ case "$(uname -m)" in
...
         unzip bin/latest-configlet.zip -d bin/
         rm bin/latest-configlet.zip
         ;;
-    (*) curl -s --location "$URL" | tar xz -C bin/ ;;
+    (*) curl --header $HEADER -s --location "$URL" | tar xz -C bin/ ;;
```
ee7:
```suggestion
    (*) curl --header "$HEADER" -s --location "$URL" | tar xz -C bin/ ;;
```

ee7 (commented):
I confirm that:
- The diff now better matches the commit message.
- `shellcheck fetch-configlet` doesn't produce any output after these suggested changes are made.
- The `fetch-configlet` style is consistent with the previous work.

I also tested that `./fetch-configlet` works on Linux, but I didn't review or test the `fetch-configlet.ps1` changes.

ErikSchierboom:
@ee7 I've tested the powershell functionality on my Windows machine.

ee7 (approved):

ee7:
@ErikSchierboom And I'm sure you know powershell much better than I do :)

ErikSchierboom:
Thanks for the review!

ee7:
Thanks for the PR!

PR #178, opened 2020-10-10
[CI] update go versions

NobbZ:
Bump the used golang versions to the 2 most recent releases.

This seems to be necessary, as #177 fails to build as the source code of the configlet uses some features that are not available in the foremerly used 1.8 and 1.9 of go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ language: go
 sudo: false
 
 go:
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

